PR #1, opened 2015-03-26
Only report gauges when they change

mpetazzoni:
For Settable{Double,Long}Gauge, only report them to SignalFx when they
their value has been set since the last time we reported them.

**Note:** I need to test this more thoroughly locally before you merge it, but I want a second (or more) pair of eyes on it too.


mpetazzoni:
Forgot some files in there, forced-push to update. Hopefully TravisCI doesn't get confused.

cep21:
oh noez your code doesn't build on JDK8

mpetazzoni:
We'll go for another solution.

PR #14, opened 2015-11-26
add events to protobuf

mdubbyap:


mdubbyap:
@jloomis i'm adding proper event support to ingest, which means it makes sense to do protobuf now as well (i think that was somewhere on your distant todo list). So far this is working for me, let me know if i screwed anything up.

signalfx-protoc/src/main/protobuf/signal_fx_protocol_buffers.proto
```diff
@@ -49,3 +48,61 @@ message PointValue {
...
 }
+
+/**
+* Different categories of events supported
+*/
+enum EventCategory {
+    /**
+     * Created by user via UI or API, e.g. a deployment event
+     */
+    USER_DEFINED = 1;
```
jloomis: make values match what we have in thrift? (ALERT = 100000 etc)

mdubbyap:
Updated.

mdubbyap:
Also note, i've added the exception category for @philatsf 's log appender.  I'll update the thrift file soon as well.

PR #18, opened 2016-03-01
removed system out print when sending metrics

AlexanderKM:


AlexanderKM:
With a high number of metrics being sent, this printing becomes way too verbose, making reading logs almost impossible.

mdubbyap:
Lgtm.

PR #21, opened 2016-03-24
Support properties on protocol buffers

cep21:


cep21:
@tedoc2000 Using existing property val

PR #22, opened 2016-05-05
Support events-only mode in AggregateMetricSender

mpetazzoni:
Adds support for "events-only" mode in the `AggregateMetricSender`. Add the appropriate constructor, and refactor the `Session.close()` method to correctly support the `dataPointReceiverFactory` being _null_.


mpetazzoni:
We'll need to release 0.0.27 after this is merged.

PR #27, opened 2016-08-10
Add signalflow support

dgriff-tools:


signalfx-java/src/main/java/com/signalfx/signalflow/Channel.java
```diff
@@ -0,0 +1,97 @@
...
+
+import java.io.Closeable;
+import java.util.Iterator;
+
+import org.apache.commons.lang3.RandomStringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Copyright (C) 2016 SignalFx, Inc. All rights reserved.
```
mpetazzoni: Copyright should be a distinct comment at the top of the file (before `package`).

signalfx-java/src/main/java/com/signalfx/signalflow/Channel.java
```diff
@@ -0,0 +1,97 @@
...
+ * 
+ * Abstract immutable representation for open channels that receive streaming data from a SignalFlow
+ * computation.
+ * 
+ * Channel objects bridge the gap between an underlying transport and a higher-level Computation
+ * object by providing a transport-agnostic and encoding-agnostic access to the stream of
+ * messages.StreamMessage objects that are received for a given computation.
+ * 
+ * Channels are iterable that return ChannelMessage instances.
```
mpetazzoni: Add `@author` below after an empty comment line.

signalfx-java/src/main/java/com/signalfx/signalflow/Channel.java
```diff
@@ -0,0 +1,97 @@
...
+ */
+public abstract class Channel implements Iterator<ChannelMessage>, Closeable {
+
+    protected static final Logger log = LoggerFactory.getLogger(Channel.class);
+    private static final int CHANNEL_NAME_LENGTH = 8;
+
+    // should be set by subclass implementation
+    protected String type = "undefined";
```
mpetazzoni: Take this in as a constructor parameter and have subclasses pass it in then. Right now, you're not enforcing that it gets set by the subclass implementation.

signalfx-java/src/main/java/com/signalfx/signalflow/ChannelMessage.java
```diff
@@ -0,0 +1,345 @@
...
+            ChannelMessage message = null;
+
+            switch (streamMessage.getKind()) {
+
+            case CONTROL:
+                @SuppressWarnings("unchecked")
+                Map<String, Object> dataObject = mapper.readValue(streamMessage.getData(),
+                        Map.class);
+
+                if (Type.STREAM_START.name().equals(dataObject.get("event"))) {
```
mpetazzoni: Hmm, you need to improve your Jackson-fu, because if you deserialize correctly you should be able to get the stream message directly as an instance of `StreamStartMessage` or `JobStartMessage`, etc, without needing this if/elseif/... block. See dynamic typing with Jackson.

For an example in Java, see `ClientMessage.java` in prototype and look at the `@JsonSubTypes` annotation.

signalfx-java/src/main/java/com/signalfx/signalflow/ComputationAbortedException.java
```diff
@@ -0,0 +1,39 @@
...
+ *  Exception thrown if the computation is aborted during its execution.
+ */
+public class ComputationAbortedException extends RuntimeException {
+
+    private static final long serialVersionUID = 1L;
+
+    protected String state;
+    protected String reason;
+
+    public ComputationAbortedException(Map<String, Object> abortInfo) {
```
mpetazzoni: The `abortInfo` should be `Map<String, String>`

signalfx-java/src/main/java/com/signalfx/signalflow/ComputationHandler.java
```diff
@@ -0,0 +1,147 @@
...
+import com.signalfx.signalflow.ChannelMessage.EventMessage;
+import com.signalfx.signalflow.ChannelMessage.JobProgressMessage;
+import com.signalfx.signalflow.ChannelMessage.JobStartMessage;
+import com.signalfx.signalflow.ChannelMessage.MetadataMessage;
+import com.signalfx.signalflow.Computation.State;
+
+/**
+ * Copyright (C) 2016 SignalFx, Inc. All rights reserved.
+ * 
+ * Class that provides basic plumbing used by subclasses to processing computation
```
mpetazzoni: Unfinished?

signalfx-java/src/main/java/com/signalfx/signalflow/ComputationHandler.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        return this.computation;
+    }
+
+    @Override
+    public void run() {
+        try {
+            call();
+        } catch (ComputationAbortedException ex) {
```
mpetazzoni: If you're providing this class as a base for people to build on, you shouldn't impose on them that all errors are ignored and logged.

signalfx-java/src/main/java/com/signalfx/signalflow/Channel.java
dgriff-tools: I have seen it done both ways .. but will make the change.

signalfx-java/src/main/java/com/signalfx/signalflow/StreamMessage.java
```diff
@@ -0,0 +1,129 @@
...
+        this.event = "message";
+        kind = Kind.INFORMATION;
+    }
+
+    public StreamMessage(String event, String id, String data) {
+        this.event = event;
+        this.id = id;
+        this.data = data;
+
+        if (isKind(Kind.CONTROL)) {
```
mpetazzoni: What happened to `Kind.valueOf(event)` ?

signalfx-java/src/main/java/com/signalfx/signalflow/StreamMessage.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    public void setData(String data) {
+        this.data = data;
+    }
+
+    public String toString() {
+        StringBuilder builder = new StringBuilder();
+
+        builder.append("{ \"event\" : ");
```
mpetazzoni: Why are you building JSON by hand?

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowTransport.java
```diff
@@ -0,0 +1,21 @@
+package com.signalfx.signalflow;
+
+import java.util.Map;
+
+/**
+ * Interface for transports to the SignalFlow API
+ */
+public interface SignalFlowTransport {
+
+    public Channel attach(String handle, Map<String, String> parameters);
```
mpetazzoni: Interface methods should always be fully documented.

signalfx-java/src/main/java/com/signalfx/signalflow/Channel.java
dgriff-tools: I don't typically add author to anything that isn't frozen for eternity .. but can do it.

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowException.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * A generic error encountered when interacting with the SignalFx SignalFlow API.
+ */
+public class SignalFlowException extends RuntimeException {
+
+    private static final long serialVersionUID = 1L;
+
+    protected Integer code;
+
+    public SignalFlowException(final Integer code) {
```
mpetazzoni: Why are you using an `Integer` object instead of a primitive `int` ?

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
```diff
@@ -0,0 +1,197 @@
...
+     * @param handle
+     *            - computation id
+     * @param filters
+     *            - filter written in signalflow language
+     * @param resolution
+     *            - Optional desired data resolution, in milliseconds
+     * @return computation instance
+     */
+    public Computation attach(String handle, String filters, Integer resolution) {
+
```
mpetazzoni: rm

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    /**
+     * Keepalive a SignalFlow computation.
+     * 
+     * @param handle
+     *            - computation id
+     */
+    public void keepalive(String handle) {
+
```
mpetazzoni: rm

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * Stop a SignalFlow computation
+     * 
+     * @param handle
+     *            - computation id
+     * @param reason
+     *            - Optional description of why stop was called
+     */
+    public void stop(String handle, String reason) {
+
```
mpetazzoni: rm

signalfx-java/src/main/java/com/signalfx/signalflow/Channel.java
dgriff-tools: type was based on python example that had tyes sse and ws.  this is an abstract class with the intention that the impl will define things like type.

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
```diff
@@ -0,0 +1,197 @@
...
+     * @param start
+     *            - Optional millisecond start timestamp
+     * @param stop
+     *            - Optional millisecond stop timestamp
+     * @param resolution
+     *            - Optional desired data resolution, in milliseconds
+     * @param maxDelay
+     *            - Optional desired maximum data delay, in milliseconds
+     */
+    public void start(String program, Long start, Long stop, Integer resolution, Integer maxDelay) {
```
mpetazzoni: primitives

signalfx-java/src/main/java/com/signalfx/signalflow/ChannelMessage.java
dgriff-tools: i think you still need to define which class type

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
```diff
@@ -0,0 +1,197 @@
...
+ */
+public class SignalFlowClient {
+
+    private SignalFlowTransport transport;
+
+    /**
+     * Client Constructor that uses default transport/settings
+     * 
+     * @param token
+     *            - user api token
```
mpetazzoni: Don't include that `-` in your parameter documentation (everywhere). This is not how Javadoc formats your stuff.

signalfx-java/src/main/java/com/signalfx/signalflow/ComputationAbortedException.java
dgriff-tools: k .. I wasn't 100% sure if it would always be string for value so sided on safety.  will make the change

signalfx-java/src/main/java/com/signalfx/signalflow/ComputationHandler.java
dgriff-tools: how so?  since it is an abstract class the subclass will implement any of the onmessage methods they want.  or did you mean the class comment was unfinished?

signalfx-java/src/main/java/com/signalfx/signalflow/ComputationHandler.java
dgriff-tools: runnable doesn't return exceptions but the callable does (a reason I have both).  so documentation could point out the differences.

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    /**
+     * helper method to populate map with non-null values 
+     * 
+     * @param params - map to populate
+     * @param key
+     * @param value
+     */
+    private void addParam(Map<String, String> params, String key, Object value) {
```
mpetazzoni: Instead of this, I'd rather see a method like this:

```java
private static Map<String, String> buildParams(Object... params) {
    Preconditions.checkArgument(params.length % 2 == 0);
    ImmutableMap.Builder<String, String> builder = new ImmutableMap.Builder<>();
    for (int i = 0 ; i < params.length ; i += 2) {
        if (params[i] != null && params[i+1] != null) {
            builder.put(params[i].toString(), params[i+1].toString());
        }
    }
    return builder.build();
}
```

Then you can just use, for example in `attach()`:

```java
public Computation attach(String handle, String filters, Integer resolution) {
    return new Computation(this.transport, handle, buildParams("filters", filters, "resolution", resolution), true);
}
```

mpetazzoni:
First batch. Great work! Please look those comments over. I'll wait until you improve the deserialization bits before continuing.

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
dgriff-tools: yep .. will change

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
dgriff-tools: ok

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
dgriff-tools: ok

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
dgriff-tools: ok

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowException.java
dgriff-tools: i like null better than a false value

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowTransport.java
dgriff-tools: np .. will document

signalfx-java/src/main/java/com/signalfx/signalflow/StreamMessage.java
dgriff-tools: i like it the hard way. actually was planning to change this anyway (forgot)

signalfx-java/src/main/java/com/signalfx/signalflow/StreamMessage.java
dgriff-tools: i can add that

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowException.java
mpetazzoni: But you should be guaranteed to have an error code at this point.

signalfx-java/src/main/java/com/signalfx/signalflow/StreamMessage.java
mpetazzoni: Then in your `Kind` enum, statically build a `Map<String, Kind>` and add a static method `findByEventName(String event)` that looks up that map.

signalfx-java/src/main/java/com/signalfx/signalflow/ComputationHandler.java
mpetazzoni: My point is that you offer no way for the implementer to receive and deal with those exceptions (not only the one line 133, but all the other catch blocks as well).

signalfx-java/src/main/java/com/signalfx/signalflow/ComputationHandler.java
mpetazzoni: "to process computation...?" what? output?

signalfx-java/src/main/java/com/signalfx/signalflow/Channel.java
mpetazzoni: In the Python library there are two transport implementations indeed, SSE and WS. But they don't keep a string type around like that. The implementation class is enough information.

signalfx-java/src/main/java/com/signalfx/signalflow/Channel.java
mpetazzoni: We actually talked about this with @mdubbyap; `@author` should be `SignalFx` since this is public open source code.

signalfx-java/src/main/java/com/signalfx/signalflow/Channel.java
mpetazzoni: Thanks; just keeping with our existing style.

signalfx-java/src/main/java/com/signalfx/signalflow/Channel.java
dgriff-tools: i removed it

signalfx-java/src/main/java/com/signalfx/signalflow/ComputationHandler.java
dgriff-tools: i am improving the documented methods .. should make sense when you see it

signalfx-java/src/main/java/com/signalfx/signalflow/ComputationHandler.java
dgriff-tools: if they execute as callable then the cause exception will be available from ExecutionException from future.get.  But to remedy this I could simply remove the runnable interface as that is the only one without a means to return exceptions to user.

signalfx-java/src/main/java/com/signalfx/signalflow/Channel.java
dgriff-tools: k .. adding author

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
```diff
@@ -0,0 +1,531 @@
...
+ * Server-Sent Events transport.
+ * 
+ * Implements a transport to the SignalFlow API that uses simple HTTP requests and reads Server-Sent
+ * Events streams back from SignalFx. One connection per SignalFlow computation is required when
+ * using this transport. This is a good transport for single, ad-hoc computations. For most use
+ * cases though, the WebSocket-based transport is more efficient and has lower latency.
+ * 
+ * @author dgriff
+ */
+public class ServerSentEventsTransport implements SignalFlowTransport {
```
mpetazzoni: Can you use Jetty's SSE client here? Or does it not support setting custom headers for authentication?

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
```diff
@@ -0,0 +1,531 @@
...
+    }
+
+    /**
+     * Builder of SSE Transport Instance
+     */
+    public static class TransportBuilder {
+
+        private String token;
+        private String protocol = "https";
+        private String host = "stream.signalfx.com";
```
mpetazzoni: Make this a constant somewhere

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
```diff
@@ -0,0 +1,531 @@
...
+
+        public TransportConnection(SignalFxEndpoint endpoint, int timeoutMs) {
+            super(endpoint, timeoutMs, new BasicHttpClientConnectionManager());
+
+            this.transportRequestConfig = RequestConfig.custom().setSocketTimeout(0)
+                    .setConnectionRequestTimeout(this.requestConfig.getConnectionRequestTimeout())
+                    .setConnectTimeout(this.requestConfig.getConnectTimeout())
+                    .setProxy(this.requestConfig.getProxy()).build();
+
+            if (log.isDebugEnabled()) {
```
mpetazzoni: You don't really need to gate your logging statements with this check (it's the first thing `log.debug()` will do), unless the methods you're calling on your parameters are expensive and you're logging in a hot path. It doesn't seem to to be the case here (and most everywhere else) so you can drop those.

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
```diff
@@ -0,0 +1,531 @@
...
+                    throw new SignalFlowException(statusLine.getStatusCode(), errorMessage);
+                }
+
+                return response;
+
+            } catch (IOException ex) {
+                throw new SignalFlowException(
+                        "failed post [ " + httpPost + " ] reason: " + ex.getMessage(), ex);
+            } catch (URISyntaxException ex) {
+                throw new SignalFlowException("failed post reason " + ex.getMessage(), ex);
```
mpetazzoni: "failed post" seems like an obscure error message. The user doesn't necessarily know (or needs to know) that a HTTP POST is going on behind the scenes.

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
```diff
@@ -0,0 +1,531 @@
...
+            this.connection = connection;
+            this.response = response;
+            this.responseHttpEntity = response.getEntity();
+            this.streamParser = new TransportEventStreamParser(
+                    this.responseHttpEntity.getContent());
+            this.iterator = this.streamParser;
+
+            if (log.isDebugEnabled()) {
+                log.debug(
+                        "constructed " + this.toString() + " of type " + this.getClass().getName());
```
mpetazzoni: Don't do string formatting yourself.

```java
log.debug("constructed {} of type {}", this, this.getClass().getName());
```

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
```diff
@@ -0,0 +1,531 @@
...
+        private static final String DATA = "data";
+        private static final String RETRY = "retry";
+        private static final String DEFAULT_EVENT = "message";
+        private static final String EMPTY_STRING = "";
+        private static final Pattern DIGITS_ONLY = Pattern.compile("^[\\d]+$");
+
+        private BufferedReader eventStreamReader;
+        private boolean endOfStreamReached = false;
+
+        private Integer reconnectionTimeout = 1000; // default is 1 second
```
mpetazzoni: Why is this an `Integer` and not an `int`? Also, consider including the unit in the variable name, like `reconnectionTimeoutMs`

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
```diff
@@ -0,0 +1,531 @@
...
+
+            while ((endOfStreamReached == false) && (eventStreamReader != null)
+                    && (nextMessage == null)) {
+                parseNext();
+            }
+
+            if (nextMessage != null) {
+                StreamMessage message = this.nextMessage;
+
+                // important to set next message to null here
```
mpetazzoni: Ok, but why?

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
```diff
@@ -0,0 +1,531 @@
...
+                try {
+                    this.eventStreamReader.close();
+                    this.eventStreamReader = null;
+                } catch (IOException ex) {
+                    log.error("failed to close event stream", ex);
+                }
+            }
+        }
+
+        private void parseNext() {
```
mpetazzoni: This is the part I wish we didn't have to do by using Jetty's SSE client.

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
```diff
@@ -0,0 +1,200 @@
...
+
+    /**
+     * Execute the given SignalFlow program and stream the output back.
+     * 
+     * @param program
+     *            computation written in signalflow language
+     * @return computation instance
+     */
+    public Computation execute(String program) {
+        return new Computation(this.transport, program, new HashMap<String, String>(), false);
```
mpetazzoni: Nit: use `Collections.emptyMap()`; avoids an allocation.

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
```diff
@@ -0,0 +1,200 @@
...
+
+    /**
+     * Start executing the given SignalFlow program without being attached to the output of the
+     * computation.
+     * 
+     * @param program
+     *            computation written in signalflow language
+     */
+    public void start(String program) {
+        this.transport.start(program, new HashMap<String, String>());
```
mpetazzoni: Same about `Collections.emptyMap()`

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
```diff
@@ -0,0 +1,200 @@
...
+     * Stop a SignalFlow computation
+     * 
+     * @param handle
+     *            computation id
+     * @param reason
+     *            Optional description of why stop was called
+     */
+    public void stop(String handle, String reason) {
+        Map<String, String> params = new HashMap<String, String>();
+        addParam(params, "reason", reason);
```
mpetazzoni: Use `buildParams()` or `ImmutableMap.of("reason", reason)`

signalfx-java/src/main/java/com/signalfx/signalflow/StreamMessage.java
```diff
@@ -0,0 +1,111 @@
...
+            this.specName = specName;
+        }
+
+        public String toString() {
+            return this.specName;
+        }
+    };
+
+    private static Map<String, Kind> SPEC_NAME_KINDS = new HashMap<String, Kind>();
+    static {
```
mpetazzoni: That's... almost what I meant, but not quite. Just add this inside the `Kind` enum above after the enum entries:

```java
private static final Map<String, Kind> KINDS = new HashMap<>();
static {
    for (Kind kind : Kind.values()) {
        KINDS.put(kind.specName, kind);
    }
}

public static Kind fromSpecName(String specName) {
    Kind kind = KINDS.get(specName);
    Preconditions.checkArgument(kind != null);
    return kind;
}
```

signalfx-java/src/main/java/com/signalfx/signalflow/StreamMessage.java
```diff
@@ -0,0 +1,111 @@
...
+            kind = Kind.INFORMATION; // default kind
+        }
+    }
+
+    public Kind getKind() {
+        return this.kind;
+    }
+
+    public boolean isKind(Kind kind) {
+        return kind.toString().equals(this.event);
```
mpetazzoni: Why is this not `return this.kind == kind` ?

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
dgriff-tools: I'll look at jetty sse and get back.  Since SSE is pretty simple I thought it would be ok to impl it and use the existing apache httpclient. But I am using jetty websocket client anyway so maybe use jetty sse would make better since.

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
dgriff-tools: ok

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
dgriff-tools: i only gate them when there is work like string concatentation in play

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
dgriff-tools: true.  I'll make more user friendly

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
dgriff-tools: ok

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
dgriff-tools: ok

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
dgriff-tools: i'll add the why

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
dgriff-tools: i'll look at the jetty sse client

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
dgriff-tools: ok .. will change

signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
dgriff-tools: k

signalfx-java/src/main/java/com/signalfx/signalflow/StreamMessage.java
dgriff-tools: ok

signalfx-java/src/main/java/com/signalfx/signalflow/StreamMessage.java
dgriff-tools: will change

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
mpetazzoni: You shouldn't be concatenating to begin with:

```java
log.debug("constructed request config: {}", this.transportRequestConfig);
```

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+
+        } catch (Exception ex) {
+            throw new SignalFlowException("failed to construct websocket transport", ex);
+        }
+    }
+
+    @Override
+    public Channel attach(String handle, final Map<String, String> parameters) {
+        if (log.isDebugEnabled()) {
+            log.debug("attach: [ " + handle + " ] with parameters: " + parameters);
```
mpetazzoni: Here again:

```java
log.debug("attach: [{}] with parameters: {}", handle, parameters);
```

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+        }
+
+        Channel channel = new TransportChannel(transportConnection);
+        HashMap<String, String> request = new HashMap<String, String>(parameters);
+        request.putAll(buildParams("type", "attach", "handle", handle));
```
mpetazzoni: If you're adding string/string pairs, avoid building the additional map that `buildParams` builds:

```java
Map<String, String> request = new HashMap<>();
request.putAll(parameters);
request.put("type", "attach");
request.put("handle", handle);
```

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+    public void close(int code, String reason) {
+        if ((transportConnection.getConnection() != null)
+                && (transportConnection.getConnection().isOpen())) {
+            transportConnection.close(code, reason);
+            try {
+                this.webSocketClient.getFactory().stop();
+            } catch (Exception ex) {
+                log.error("error while stopping websocketfactory", ex);
+            }
+            log.info("transport closed");
```
mpetazzoni: debug level?

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+    public void keepalive(String handle) {
+        if (log.isDebugEnabled()) {
+            log.debug("keepalive: [ " + handle + " ]");
+        }
+
+        Map<String, String> request = buildParams("type", "keepalive", "handle", handle);
+        transportConnection.sendMessage(request);
+    }
+
+    private static Map<String, String> buildParams(Object... params) {
```
mpetazzoni: If you still use this here given my comment above, don't have it twice between the Websocket transport and the SSE transport, but that in some shared class.

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+    /**
+     * Value Object that handles binary data message conversion in construction
+     */
+    protected static class TransportDataMessage {
+
+        protected byte version;
+        protected byte messageType;
+        protected String channelName;
+        protected long logicalTimestampMs;
+        protected ArrayList<Map<String, Object>> data = new ArrayList<Map<String, Object>>();
```
mpetazzoni: Always use interface types in the variable definition. No need to specify the generic type in the variable assignment, either:

```java
protected List<Map<String, Object>> data = new ArrayList<>();
```

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+
+        public TransportDataMessage(byte[] data) {
+            try {
+                this.version = data[0];
+                this.messageType = data[1];
+                this.channelName = new String(data, 4, 16, "UTF-8");
+
+                if (messageType == 5) {
+                    byte[] logicalTimestampMsBytes = Arrays.copyOfRange(data, 20, 28);
+                    this.logicalTimestampMs = ByteBuffer.wrap(logicalTimestampMsBytes).getLong();
```
mpetazzoni: Don't wrap individual sections of the data byte array into `ByteBuffer`s to decode them. Wrap the whole thing, and then navigate the `ByteBuffer` to read sections of it.

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+                } else {
+                    log.warn("Unsupported binary message type {}", messageType);
+                }
+            } catch (Exception ex) {
+                log.error("failed to construct transport data message", ex);
+            }
+        }
+
+        public int getVersion() {
+            return (int) version;
```
mpetazzoni: If you have to cast here then you probably should have casted when reading the value from the data array.

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+        }
+
+        public String getChannelName() {
+            return channelName;
+        }
+
+        public int getMessageType() {
+            return (int) messageType;
```
mpetazzoni: This should return a `Kind`. Which means you need to make sure your `Kind` enum matches the naming and ordering of `StreamMessage.Type` (in SignalBoost).

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+        }
+
+        @Override
+        public void onMessage(byte[] data, int offset, int length) {
+
+            try {
+                byte[] messageBytes = Arrays.copyOfRange(data, offset, offset + length);
+                TransportDataMessage dataMessage = new TransportDataMessage(messageBytes);
+
+                if (dataMessage.getMessageType() == 5) {
```
mpetazzoni: Use enum type

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+            Map<String, String> authRequest = new HashMap<String, String>();
+            authRequest.put("type", "authenticate");
+            authRequest.put("token", this.token);
+
+            sendMessage(authRequest);
+        }
+
+        @Override
+        public void onMessage(byte[] data, int offset, int length) {
+
```
mpetazzoni: rm

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+            try {
+                byte[] messageBytes = Arrays.copyOfRange(data, offset, offset + length);
```
mpetazzoni: I'd encourage you to use a ByteBuffer instead so can be zero copy:

```java
ByteBuffer message = ByteBuffer.wrap(data);
message.position(offset);
message.limit(offset + length);
```

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+                    }
+                }
+            } catch (JsonProcessingException ex) {
+                log.error("failed to process messages", ex);
+            }
+        }
+
+        @Override
+        public void onMessage(String data) {
+
```
mpetazzoni: rm

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+            try {
+                @SuppressWarnings("unchecked")
+                Map<String, Object> dataMap = objectMapper.readValue(data, Map.class);
+
+                // Intercept KEEP_ALIVE messages
+                String event = (String) dataMap.get("event");
+                if ((event != null) && (event.equals("KEEP_ALIVE"))) {
```
mpetazzoni: Don't need the inner parentheses; also, simpler to do: `"KEEP_ALIVE".equals(event)`.

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+                    return;
+                }
+
+                event = (String) dataMap.get("type");
```
mpetazzoni: That's confusing

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+                if (event != null) {
```
mpetazzoni: Prefer early-return like above

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+                    log.debug("ignoring message. {}", dataMap);
+                }
+
+            } catch (IOException ex) {
+                log.error("failed to process messages", ex);
+            }
+        }
+
+        public void sendMessage(final Map<String, String> request) {
+
```
mpetazzoni: rm

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,592 @@
...
+            try {
+                String message = objectMapper.writeValueAsString(request);
+                this.connection.sendMessage(message);
+            } catch (IOException ex) {
+                throw new SignalFlowException("failed to send message", ex);
```
mpetazzoni: Should this cause the transport to close?

README.md
```diff
@@ -418,3 +418,59 @@ Example with directive to bypass proxy for localhost and host.mydomain.com
...
+* `SignalFlow Overview`_
+* `Getting started with the SignalFlow API`_
+
+.. _SignalFlow Overview: https://developers.signalfx.com/docs/signalflow-overview
+.. _Getting started with the SignalFlow API: https://developers.signalfx.com/docs/getting-started-with-the-signalflow-api
+
+Executing a SignalFlow program is very simple with this client library:
+
+```java
+
```
mpetazzoni: rm empty line, and don't need the first-level indent of the entire block of code below. Same applies to other block line 468

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: i would have done it that way but you had such a fondness for buildparams i let it go

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: sure

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
dgriff-tools: fixed all the log messages to not concat and format via log impl. then removed the debug check

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: fixed

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: done

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: removed as not needed

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: oversight .. thanks

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: done but in a way that is easy to read.

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: no cast needed. removed

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: done.  a little of overkill but done

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: done.  still a little overkill but done

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: done but my preference is to give a line under method signature for better viewing.
you obviously have a different opinion.

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: not doing this as I like the way it works now.

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: done

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: done

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: changed to type

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: once upon a time early returns like that were considered poor programming.
anyway .. done

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: done

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: if it doesn't throw exception then the state of transport or channel is indeterminate.

signalfx-java/src/main/java/com/signalfx/signalflow/ServerSentEventsTransport.java
dgriff-tools: where do you see the jetty client for sse?  I see eventsourceservlet for jetty 9 (later than the jetty 8 with java 1.6 support) but that is not useful for a java client perspective.

The impl in this PR uses the existing apache httpclient with proxy support and should implement the sse spec correctly and is pretty minimal is size.

mpetazzoni: My bad, I meant Jersey. I always get those two mixed up.

See https://jersey.java.net/documentation/latest/sse.html#d0e11970

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
mpetazzoni: That's sort of my point. I don't think we should ever leave the channel is an indeterminate state. Either it's open and usable, or it's closed.

README.md
```diff
@@ -418,3 +418,54 @@ Example with directive to bypass proxy for localhost and host.mydomain.com
...
+                    eventMessage.getTimestampMs() + ": " + eventMessage.getProperties());
+            break;
+        }
+    }
+} finally {
+    flow.close();
+}
+```
+
+Metadata about the timeseries is received from the iterable stream, and
```
mpetazzoni: `s/stream, and is automatically/stream. It is also automatically/`

signalfx-java/src/main/java/com/signalfx/signalflow/StreamMessage.java
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        public static Kind fromSpecName(String specName) {
+            Kind kind = SPECNAME_KINDS.get(specName);
+            Preconditions.checkArgument(kind != null);
+            return kind;
+        }
+        
+        public static Kind fromBinaryType(int binaryType) {
+            Kind kind = BINARYTYPE_KINDS.get(new Integer(binaryType));
```
mpetazzoni: You don't need the `new Integer`, Java will autobox.

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,571 @@
...
+
+                this.version = buffer.get(0);
+                this.kind = Kind.fromBinaryType(buffer.get(1));
+                this.channelName = new String(data, 4, 16, "UTF-8");
+
+                if (this.kind == Kind.DATA) {
+                    this.logicalTimestampMs = buffer.getLong(20);
+
+                    byte[] payload = Arrays.copyOfRange(data, 32, data.length);
+                    buffer = ByteBuffer.wrap(payload);
```
mpetazzoni: Why are you wrapping again? You should be able to use the original buffer, just moving your position around.

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -0,0 +1,571 @@
...
+        }
+
+        public TransportConnection(String token) {
+            this.token = token;
+        }
+
+        @Override
+        public void onClose(int code, String reason) {
+            if (log.isDebugEnabled()) {
+                log.debug("websocket connection closed - code: " + code + " reason: " + reason);
```
mpetazzoni: Use logger formatting:

```java
log.debug("websocket connection closed ({} {})", code, reason);
```

signalfx-java/src/main/java/com/signalfx/signalflow/StreamMessage.java
```diff
@@ -0,0 +1,130 @@
...
+        kind = Kind.INFORMATION;
+    }
+
+    public StreamMessage(String event, String id, String data) {
+        this.event = event;
+        this.id = id;
+        this.data = data;
+
+        kind = Kind.fromSpecName(event);
+        if (kind == null) {
```
mpetazzoni: This can't happen given how `fromSpecName()` is implemented.

signalfx-java/src/main/java/com/signalfx/signalflow/StreamMessage.java
```diff
@@ -0,0 +1,130 @@
...
+        public String toString() {
+            return this.specName;
+        }
+
+        private static final Map<String, Kind> SPECNAME_KINDS = new HashMap<String, Kind>();
+        private static final Map<Integer, Kind> BINARYTYPE_KINDS = new HashMap<Integer, Kind>();
+        static {
+            for (Kind kind : Kind.values()) {
+                SPECNAME_KINDS.put(kind.specName, kind);
+                BINARYTYPE_KINDS.put(new Integer(kind.getBinaryType()), kind);
```
mpetazzoni: No need for explicit boxing

mpetazzoni:
Looks pretty good. There's a few remaining nits, especially around logging (make sure you use slf4j log formatting on all your logging statements and remove the gating if blocks around if they're not needed anymore).

Fix that, plus the few other notes, make a clean commit message, and we can merge. Let's do the Jersey SSE stuff in another PR after this one so we can move forward.

Thanks!

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
dgriff-tools: the original caller here would always be the SignalFlowClient class (even on transport initial construction).  So the thrown runtime exception should prevent any transport or channel from existing.  And also be the correct response on other actions like stopping computation.  But I could force a complete transport close on any exception .. but I am not sure if that is the correct thing to do for all IOExceptions from the sendmessage.

PR #32, opened 2016-11-15
Added method (convertDurationsTo) to convert default reporting units (nanoseconds) to user…

harnitsignalfx:
…-defined Time Unit


harnitsignalfx:
Let me know if this doesn't fit the SFx standards or I need to change a few (or more) things

mpetazzoni:
I'm 👎 on this.

I understand why people think they want this, but really they shouldn't.

I'm afraid this could cause much more harm than good, especially if multiple emitters don't report the same metrics (with different dimensions) using the same units. It becomes impossible to do any kind of computation across those timeseries because they'll have values in different orders of magnitude.

Time units are like timezones. You want to measure and store in the most absolute form possible (nanoseconds for durations, UNIX timestamps for time), and then have the UI present this data in whatever unit (or timezone) the user prefers most.

harnitsignalfx:
Hmm, I see your point. From the point of view of the customer though, if their value is far too granular to provide value as is, the only alternative left for them is to use Scale on every one of their signalflow jobs within every single chart. 
Also, it looks like the duration and unit are available in the default codahale reporter. I'll be making a few changes that Ted mentioned on making this change equivalent to how its typically done by the original reporter.
Having said that - if you feel very strongly about it, then I can close this PR.

PR #34, opened 2016-12-01
Support expired-tsid stream message

dgriff-tools:
New stream message type expired-tsid tells us about timeseries that are
no longer interesting in the computation (if they do re-appear, we'll
get new metadata for them). We use this message to remove entries from
the computation object's internal metadata storage map.

### DIFF
--- a/signalfx-java/src/main/java/com/signalfx/signalflow/Computation.java
+++ b/signalfx-java/src/main/java/com/signalfx/signalflow/Computation.java
@@ -16,6 +16,7 @@ import java.util.NoSuchElementException;
 import com.signalfx.signalflow.ChannelMessage.ChannelAbortMessage;
 import com.signalfx.signalflow.ChannelMessage.DataMessage;
 import com.signalfx.signalflow.ChannelMessage.ErrorMessage;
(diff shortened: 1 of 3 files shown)


signalfx-java/src/main/java/com/signalfx/signalflow/Computation.java
```diff
@@ -268,13 +269,20 @@ private void parseNext()
...
                     MetadataMessage metadataMessage = (MetadataMessage) message;
                     metadata.put(metadataMessage.getTsId(), metadataMessage.getProperties());
                     this.nextMessage = message;
                     break;
 
+                case EXPIRED_TSID_MESSAGE:
+                    // Intercept expired-tsid messages to clean it up.
+                    ExpiredTsIdMessage expiredTsIdMessage = (ExpiredTsIdMessage) message;
+                    if (this.metadata.get(expiredTsIdMessage.getTsId()) != null) {
+                        this.metadata.remove(expiredTsIdMessage.getTsId());
```
mpetazzoni: Just call remove unconditionally, no need for two hash lookups.

PR #35, opened 2016-12-02
fix the version update to work correctly with child projects

dgriff-tools:
The update version script was iterating through the child projects directly but using the update child directive which caused it to fail.  This change fixes that issue and cleans up the script.


mpetazzoni:
This so doesn't need to be Python, but ok.

PR #37, opened 2017-02-03
Version 0.0.31 - Add Agent event category to protobuf

charliesignalfx:


signalfx-protoc/src/main/protobuf/signal_fx_protocol_buffers.proto
```diff
@@ -87,6 +87,10 @@ enum EventCategory {
      * Created by exception appenders to denote exceptional events
```
mdubbyap: Even though you can't actually deprecate a particular enum, i would add a comment that this is deprecated in favour of AGENT.

charliesignalfx:
@mdubbyap I removed the pom file updates because Dgriff pointed out that our commit history had changes Pulled in from branches, but the actual version changes were straight to master.  So I backed out the pom file changes.

signalfx-protoc/src/main/protobuf/signal_fx_protocol_buffers.proto
charliesignalfx: I have addressed this in the comment above ^

PR #40, opened 2017-03-21
Support (and request) compressed WebSocket messages

mpetazzoni:


signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -357,35 +302,149 @@ public void onOpen(Connection connection) {
...
+                log.error("ignoring message with unsupported type {}", type);
+                return;
+            }
 
-                if (dataMessage.getKind() == Kind.DATA) {
-                    LinkedHashMap<String, Object> dataMap = new LinkedHashMap<String, Object>();
-                    dataMap.put("logicalTimestampMs", dataMessage.getLogicalTimestampMs());
-                    dataMap.put("data", dataMessage.getData());
+            // Channel name is the 16 bytes following the binary preamble in the header.
+            String channelName = new String(data, offset + BINARY_PREAMBLE_LENGTH,
```
tedoc2000: add charset here or do we require ascii channel names?

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -357,35 +302,149 @@ public void onOpen(Connection connection) {
...
             }
         }
 
         @Override
         public void onMessage(String data) {
             try {
-                @SuppressWarnings("unchecked")
-                Map<String, Object> dataMap = objectMapper.readValue(data, Map.class);
+                Map<String, Object> dataMap = objectMapper.readValue(data,
+                        new TypeReference<Map<String, Object>>() {});
```
tedoc2000: No need to new each time. make it a static

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
mpetazzoni: Yeah it's ASCII.

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
mpetazzoni: Ack

signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
tedoc2000: better to be explicit then with new String(..., "ASCII")

mpetazzoni:
Addressed feedback, also fixed some Javadoc warnings.

PR #42, opened 2017-04-21
Correctly encode TSIDs as Base64Url strings when decoding binary data messages

mpetazzoni:
Otherwise the TSID of datapoints doesn't matchup with metadata TSIDs.


molner:
LGTM

PR #43, opened 2017-06-29
Add resetting histogram using a reservoir that resets on snapshot

mpetazzoni:


harnitsignalfx:
Thanks Max!
Since the super class seems to also support constructors that take in size of the reservoir, alpha and clock - did you want to have other constructors for the ResettingHistogram class that could take in size, alpha and clock parameters and instantiate the appropriate Reservoir?

signalfx-codahale/src/main/java/com/codahale/metrics/ResettingExponentiallyDecayingReservoir.java
```diff
@@ -0,0 +1,212 @@
...
+        }
+    }
+
+    @Override
+    public Snapshot getSnapshot() {
+        rescaleIfNeeded();
+        lockForRegularUsage();
+        try {
+            Snapshot snapshot = new WeightedSnapshot(values.values());
+            values.clear();
```
mdubbyap: should reset the count too i think.

mpetazzoni: Great catch. Updated.

PR #44, opened 2017-07-24
Take write lock instead of read lock when resetting after snapshot

harnitsignalfx:


harnitsignalfx:
@mpetazzoni - is this what you needed?

signalfx-codahale/src/main/java/com/codahale/metrics/ResettingExponentiallyDecayingReservoir.java
```diff
@@ -136,14 +136,14 @@ private void rescaleIfNeeded() {
     @Override
     public Snapshot getSnapshot() {
         rescaleIfNeeded();
-        lockForRegularUsage();
+        lockForRescale();
```
mpetazzoni: Actually, to avoid taking and releasing the lock twice, with someone else potentially taking it in the middle, can you `lockForRescale()` before `rescaleIfNeeded()`? `rescale()` will take the (re-entrant) lock, which will then be a no-op since we already own it.

mpetazzoni:
Also update your commit message to "Take write lock instead of read lock when resetting after snapshot"

harnitsignalfx:
@mpetazzoni done, thanks

PR #45, opened 2017-07-26
Emit number of samples in histogram snapshot being reported

harnitsignalfx:
@mpetazzoni to review changes with @mdubbyap


mpetazzoni:
@mdubbyap Thinking more about this, why would `.samples` have a different value than the delta of `.count` ?

mpetazzoni:
As discussed, won't be needing this.

PR #46, opened 2017-09-05
Adding metric group name as dimension to yammer reporter

keitwb:


signalfx-yammer/src/main/java/com/signalfx/codahale/reporter/AggregateMetricSenderSessionWrapper.java
```diff
@@ -241,6 +241,11 @@ void addMetric(Metric metric, MetricName codahaleName,
             }
         }
 
+        if (codahaleName.getGroup() != null && codahaleName.getGroup() != "") {
```
tedoc2000: if (!StringUtils.isEmpty(codahaleName.getGroup()) {

keitwb:
@mpetazzoni This is needed because we have a customer who has an existing application that reports metrics using yammer that reuses metric names across groups.  They want to be able to easily distinguish those metrics in the app without having to change the yammer code in their application.

keitwb:
@tedoc2000 does this mean going to 0.1.0 since our current version is 0.0.36?

tedoc2000:
@keitwb It would be a major change in functionality so I would say 1.0.0

mpetazzoni:
Are metric groups a thing in codahale too? Are we going to need this in `signalfx-codahale`? If we do, how many other customers is this going to impact (in addition to us)?

keitwb:
@mpetazzoni From what I can tell from http://metrics.dropwizard.io/3.1.0/manual/core/ it doesn't look like newer versions of Codahale has the concept of metric group names, but I'm not terribly familiar with this library, so perhaps @tedoc2000 knows.

signalfx-yammer/pom.xml
```diff
@@ -57,6 +57,10 @@
       <artifactId>guava</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.commons</groupId>
```
tedoc2000: There is an equivalent of StringUtils.isEmpty in guava.. Lets use that one and not add yet another dependency

tedoc2000:
@keitwb don't know anything about the metric_group thing, by my googling matches yours so it seems like a yammer thing only

mpetazzoni:
LGTM. @keitwb if you have some cycles can you also figure out what's going on with the TravisCI build?

PR #48, opened 2017-09-08
Turns out we need metric type in yammer too

keitwb:


tedoc2000:
Can you fix the commit message to be more information and not a statement about need. Something like:
Add "metric_type" as a dimension on yammer metrics when xxx flag is set.

Also renamed sendGroupNameAsDimension to sendExtraMetricDimensions

PR #50, opened 2017-10-23
Upgrading protobuf to 3.4.0 to match sf-base

keitwb:
Protobuf schemas will still be compiled as v2 and will be compatible with v2 parsers

### DIFF
--- a/signalfx-protoc/pom.xml
+++ b/signalfx-protoc/pom.xml
@@ -54,11 +54,16 @@
   <build>
     <plugins>
       <plugin>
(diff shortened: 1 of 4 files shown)


signalfx-protoc/pom.xml
```diff
@@ -54,11 +54,16 @@
   <build>
     <plugins>
       <plugin>
-        <groupId>com.comoyo.maven.plugins</groupId>
```
keitwb: This plugin didn't support anything over 3.0.0.

PR #51, opened 2017-11-09
Adding option to increase max binary message size

jefersonm:
Using SignalFlow with WebSocket transport and I faced some scenarios where the size of the binary information that was returning from SignalFx was bigger than the currently supported limit.

In order to fix that, an option to increase the default size was added.

### DIFF
--- a/signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
+++ b/signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
@@ -62,7 +62,7 @@ public class WebSocketTransport implements SignalFlowTransport {
     protected TransportConnection transportConnection;
 
     protected WebSocketTransport(String token, SignalFxEndpoint endpoint, int apiVersion,
-                                 int timeout, boolean compress) {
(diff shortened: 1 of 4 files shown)


signalfx-java/src/main/java/com/signalfx/signalflow/WebSocketTransport.java
```diff
@@ -192,6 +191,7 @@ public void keepalive(String handle) {
         private int timeout = DEFAULT_TIMEOUT;
         private int version = 2;
         private boolean compress = true;
+        private int maxBinaryMessageSize = -1;
```
mpetazzoni: What's the impact of using -1 as the default?

jefersonm: Hi @mpetazzoni, the current implementation is already using -1 as default parameter.

The Jetty WebSocketClient that is being used by SignalFx has -1 as default:

https://github.com/eclipse/jetty.project/blob/jetty-8.1.x/jetty-websocket/src/main/java/org/eclipse/jetty/websocket/WebSocketClient.java#L90

PR #53, opened 2018-01-17
Shaded dependencies

mpetazzoni:


pom.xml
```diff
@@ -299,6 +299,53 @@
...
+              <pattern>org.apache</pattern>
+              <shadedPattern>com.signalfx.shaded.apache</shadedPattern>
+            </relocation>
+            <relocation>
+              <pattern>com.yammer</pattern>
+              <shadedPattern>com.signalfx.shaded.yammer</shadedPattern>
+            </relocation>
+            <relocation>
+              <pattern>com.codahale</pattern>
+              <shadedPattern>com.signalfx.shaded.codahale</shadedPattern>
```
jkschneider: I'm not sure you can shade codahale and yammer, as those modules appear to have user facing API surface area with types in those modules.

Simplified example from your tests:

```
MetricRegistry metricRegistery = new MetricRegistry();
        SignalFxReporter reporter = new SignalFxReporter.Builder(metricRegistery, new StaticAuthToken(""), "myserver").build();
```

The user needs to be able to interact with `SignalFxReporter.Builder` by passing a `MetricRegistry` instance they already have, not the shaded type.

I may not be understanding the purpose of these modules though. If so, ignore me :)

jkschneider: The reason I called out Guava and Jackson for shading initially is (1) they are internal implementation details and (2) likely to cause downstream pain.

mpetazzoni: Good points, I don't think I should be shading those. All other "internal" dependencies should still be shaded though. I'll send an updated diff.

PR #56, opened 2018-03-20
Send compressed datapoints payload by default

mpetazzoni:
Unless the com.signalfx.public.java.disableHttpCompression property is
set to true, the signalfx-java library will now send GZip-compressed
datapoint payloads to the SignalFx ingest endpoint.

### DIFF
--- a/signalfx-java/src/main/java/com/signalfx/metrics/connection/AbstractHttpDataPointProtobufReceiverConnection.java
+++ b/signalfx-java/src/main/java/com/signalfx/metrics/connection/AbstractHttpDataPointProtobufReceiverConnection.java
@@ -20,10 +20,13 @@ public abstract class AbstractHttpDataPointProtobufReceiverConnection extends Ab
 
     protected static final ContentType PROTO_TYPE = ContentType.create("application/x-protobuf");
 
-    public AbstractHttpDataPointProtobufReceiverConnection(
-            SignalFxReceiverEndpoint endpoint,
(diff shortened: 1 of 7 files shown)


signalfx-java/src/main/java/com/signalfx/metrics/connection/AbstractHttpDataPointProtobufReceiverConnection.java
```diff
@@ -36,8 +36,13 @@ public void addDataPoints(String auth, List<SignalFxProtocolBuffers.DataPoint> d
             CloseableHttpResponse resp = null;
             try {
                 resp = postToEndpoint(auth, getEntityForVersion(dataPoints),
-                        getEndpointForAddDatapoints());
+                        getEndpointForAddDatapoints(),
+                        !Boolean.getBoolean(DISABLE_COMPRESSION_PROPERTY));
```
tedoc2000: figure out boolean once?

mpetazzoni: Done

mpetazzoni:
@tedoc2000 I've also fixed the error reporting with this diff. When sending datapoints, the library was not checking the HTTP status code and would never trigger the `OnSendErrorHandler` if the underlying HTTP POST did not succeed (it would only do it if it threw an exception, but not if it got a non-200 answer).

PR #58, opened 2018-04-04
Use the STRICT_HOSTNAME_VERIFIER to validate SSL certificates' hostnames

mpetazzoni:
The default hostname verifier isn't strict about subdomains, which we
should be for improved security.


molner:
LGTM

PR #63, opened 2018-12-06
Add timeout before SSL handshake

jhaber:
This patch attempts to mitigate a failure scenario we saw today which caused an outage of our SignalFx setup. Metrics stopped flowing to SignalFx, and when we took a thread dump of the process responsible for sending metrics to SignalFx, we saw that it was stuck inside of `sun.security.ssl.SSLSocketImpl.startHandshake`. Despite [already setting timeouts](https://github.com/signalfx/signalfx-java/blob/6c708b3285a286cd037853c90cba9e269157ac5f/signalfx-java/src/main/java/com/signalfx/connection/AbstractHttpReceiverConnection.java#L50-L52) six ways to Sunday, it appears that none of these apply to the SSL handshake. So we were getting stuck inside of the SSL handshake forever, at which point we had to kill the process to recover (until it happened again).

We've actually seen this failure scenario before and patched our internal HTTP client, so this PR just ports the mitigation we've used successfully for a while. First, it sets a `SocketConfig` with `SO_TIMEOUT` applied. Second, it uses a subclass of `SSLConnectionSocketFactory` in order to override the `prepareSocket` method and make double-sure that `SO_TIMEOUT` is set.

We've already applied this change internally and it resolved our outage.

/cc @szabowexler

### DIFF
--- a/signalfx-java/src/main/java/com/signalfx/metrics/connection/HttpClientConnectionManagerFactory.java
+++ b/signalfx-java/src/main/java/com/signalfx/metrics/connection/HttpClientConnectionManagerFactory.java
@@ -1,0 +1,51 @@
+package com.signalfx.metrics.connection;
+
+import java.io.IOException;
+
+import javax.net.ssl.SSLSocket;
+
+import org.apache.http.config.RegistryBuilder;
+import org.apache.http.config.SocketConfig;
+import org.apache.http.conn.HttpClientConnectionManager;
+import org.apache.http.conn.socket.ConnectionSocketFactory;
+import org.apache.http.conn.socket.PlainConnectionSocketFactory;
+import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
+import org.apache.http.impl.conn.BasicHttpClientConnectionManager;
+import org.apache.http.ssl.SSLContexts;
+
+public class HttpClientConnectionManagerFactory {
+
+  private HttpClientConnectionManagerFactory() {
+    // prevent instantiation
+  }
+
+  public static HttpClientConnectionManager withTimeoutMs(int timeoutMs) {
+    BasicHttpClientConnectionManager httpClientConnectionManager = new BasicHttpClientConnectionManager(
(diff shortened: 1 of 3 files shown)


jhaber:
This PR doesn't add any tests for this particular failure scenario as it requires a bit of work to reproduce, but let me know if you'd prefer that I try to add a test

jhaber:
Thanks for the feedback. I was expecting usages to look like:
```java
DataPointReceiver dataPointReceiver = new HttpDataPointProtobufReceiverFactory(endpoint)
    .setVersion(version)
    .setTimeoutMs(timeout)
    .createDataPointReceiver();
```

But looking through the code I see that it is actually common to use a form like:
```java
DataPointReceiverFactory dataPointReceiverFactory = new HttpDataPointProtobufReceiverFactory(endpoint)
    .setVersion(version)
    .setTimeoutMs(timeout);

// .... later ....

// this can get called repeatedly
DataPointReceiver dataPointReceiver = dataPointReceiverFactory.createDataPointReceiver();
```

So I think you're right that it's worth considering memoization of the `HttpClientConnectionManager`. I've updated the PR to match the previous behavior more closely. Now `httpClientConnectionManager` is initialized in the constructor using the default timeout, and calls to `setTimeoutMs` will update the `httpClientConnectionManager`. 

With regard to thread-safety, it doesn't appear that these classes were designed to be thread-safe (`timeoutMs`/`version` are just regular `int` fields, rather than volatile or `AtomicInteger`). However, sharing these objects across threads shouldn't cause any major problems (writes from another thread might not be visible, but you shouldn't observe it in an invalid state and you shouldn't get any exceptions). It's also worth noting that the `BasicHttpClientConnectionManager` isn't meant to be shared across threads (although for performance reasons, not correctness). From the javadoc:
> A connection manager for a single connection. This connection manager maintains only one active connection. Even though this class is fully thread-safe it ought to be used by one execution thread only, as only one thread a time can lease the connection at a time.

PR #66, opened 2019-02-11
Removing unused/experimental datapoint attribute 'Property'

jgheewala:


mpetazzoni:
Are we sure of the +1/-1 compatibility on this? If people update to a new version of `signalfx-java` that doesn't send the `Property` field, will our ingest correctly accept this data? Conversely, if our ingest stops "knowing" about this property, will it correctly accept incoming ProtoBuf data that still sends it?

jgheewala:
i tested the scenarios and didn't find any issues related to +1/-1 during my testing.

PR #70, opened 2019-02-22
Deserialize program errors correctly

cory-signalfx:
# Summary

Adjust deserialization (via object mapper) of SignalFlow errors.

# Synopsis

INT-497 reports that syntax errors do not property emit an error message, instead doing a:

```
java.util.concurrent.CompletionException: com.signalfx.signalflow.ComputationFailedException: Computation failed (null)
```

This stems from the returned JSON for an error not containing an `errors` attribute. It looks like:

```
{
  "channel" : "channel-UyP2eyq8",
  "context" : {
    "column" : 154,
    "line" : 1,
    "message" : "'index' argument cannot be greater than 'total' argument",
    "programText" : "data10 = data('*', filter=filter('deployable', 'KeplerFilteredAlertsToSlackUsingRules-sfx-alert-slack-notifier') and filter('user_instance', 'false') and partition_filter(10, 5), extrapolation='null', maxExtrapolations=-1).publish(label='series10')",
    "traceback" : [ {
      "file" : "",
      "line" : 1,
      "namespace" : "<module>",
      "text" : "data10 = data('*', filter=filter('deployable', 'KeplerFilteredAlertsToSlackUsingRules-sfx-alert-slack-notifier') and filter('user_instance', 'false') and partition_filter(10, 5), extrapolation='null', maxExtrapolations=-1).publish(label='series10')"
    } ]
  },
  "error" : 400,
  "errorType" : "ANALYTICS_PROGRAM_TYPE_ERROR",
  "message" : "Error executing SignalFlow program (ref: D0BUgZxAYAA): [File \"\", line 1, in \ndata10 = data('*', filter=filter('deployable', 'KeplerFilteredAlertsToSlackUsingRules-sfx-alert-slack-notifier') and filter('user_instance', 'false') and partition_filter(10, 5), extrapolation='null', maxExtrapolations=-1).publish(label='series10')\nTypeError: 'index' argument cannot be greater than 'total' argument]",
  "type" : "error"
}
```

The context here seems interesting, but the important bit is the message. To that end I changed the `ComputationFailedException` to have a single `String message` instead of an `ArrayList<Object>`. 

# Notes

It's not clear to me if the existing code is correct in some *other* case, and this is an exception. (No pun intended.) I can't find any documentation [here](https://developers.signalfx.com/signalflow_analytics/rest_api_messages/signalflow_messages_overview.html) that explains the contents of these error messages to validate my work.  Instead, I'm using a test case that includes a syntax error (seen above) to exercise this.

### DIFF
--- a/signalfx-java/src/main/java/com/signalfx/signalflow/ChannelMessage.java
+++ b/signalfx-java/src/main/java/com/signalfx/signalflow/ChannelMessage.java
@@ -405,13 +405,26 @@ public abstract class ChannelMessage {
 
     /**
      * Message received when the computation encounters errors during its initialization.
+     * Because the error that is returned might have one of two sets of contents, this has both an
+     * `errors` and a `message` accessor. We'll be liberal here so the code can decide how to show
+     * this error later. Optimally, these should be different message types, but that's a lot of
+     * work to handle on both the backend and client side for +1/-1.
      */
     public static class ErrorMessage extends ChannelMessage {
 
+        protected int error;
         protected ArrayList<Object> errors;
+        protected String message;
 
         public ErrorMessage() {
             this.channelMessageType = Type.ERROR_MESSAGE;
+        }
+
+        /**
+         * @return The error number, akin to an HTTP error code.
+         */
+        public int getError() {
+            return this.error;
         }
 
         /**
@@ -420,6 +433,13 @@ public abstract class ChannelMessage {
          */
         public List<Object> getErrors() {
             return this.errors;
+        }
+
+        /**
+         * @return The error message for the failure
+         */
+        public String getMessage() {
+            return this.message;
         }
     }
 }
--- a/signalfx-java/src/main/java/com/signalfx/signalflow/Computation.java
+++ b/signalfx-java/src/main/java/com/signalfx/signalflow/Computation.java
@@ -123,7 +123,7 @@ public class Computation implements Iterable<ChannelMessage>, Iterator<ChannelMe
 
     @Override
     public boolean hasNext() throws ComputationAbortedException,
-           ComputationFailedException, SignalFlowException {
+           ComputationFailedException, SignalFlowException, StreamRequestException {
         while ((state != State.STATE_COMPLETED) && (!channel.isClosed) && (nextMessage == null)) {
             parseNext();
         }
@@ -305,7 +305,15 @@ public class Computation implements Iterable<ChannelMessage>, Iterator<ChannelMe
 
                 case ERROR_MESSAGE:
                     ErrorMessage errorMessage = (ErrorMessage) message;
-                    throw new ComputationFailedException(errorMessage.getErrors());
+                    /* This is a hack based on the fact that the API can return type different
+                     * error messages with the same type. We have to check attributes to know
+                     * which error we're working with.
+                     */
+                    if (errorMessage.getMessage() != null) {
+                        throw new StreamRequestException(errorMessage.getError(), errorMessage.getMessage());
+                    } else {
+                        throw new ComputationFailedException(errorMessage.getErrors());
+                    }
                 }
             }
 
--- a/signalfx-java/src/main/java/com/signalfx/signalflow/StreamRequestException.java
+++ b/signalfx-java/src/main/java/com/signalfx/signalflow/StreamRequestException.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (C) 2019 SignalFx, Inc. All rights reserved.
+ */
+package com.signalfx.signalflow;
+
+import java.util.List;
+
+/**
+ * Exception thrown when the computation fails at request time, possibly for syntax errors.
+ *
+ * @author cwatson
+ */
+public class StreamRequestException extends RuntimeException {
+
+    private static final long serialVersionUID = 1L;
+
+    protected int errorCode;
+    protected String message;
+
+    public StreamRequestException(int errorCode, String message) {
+        super("Computation failed (" + message + ") code: " + errorCode);
+        this.message = message;
+    }
+
+    public int getErrorCode() {
+        return this.errorCode;
+    }
+
+    public String getMessage() {
+        return this.message;
+    }
+}


mpetazzoni:
See https://github.com/signalfx/prototype/blob/master/stream/lib/framework/src/main/java/sf/stream/StreamMessage.java#L344 and https://github.com/signalfx/prototype/blob/master/stream/lib/framework/src/main/java/sf/stream/StreamMessage.java#L408

cory-signalfx:
Alrighty, @mpetazzoni. This takes a much more liberal approach of just allowing either form of error. We decipher it in the body and move along to avoid having to make changes to the backend.

cory-signalfx:
Ping @mpetazzoni!

signalfx-java/src/main/java/com/signalfx/signalflow/StreamRequestException.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (C) 2016 SignalFx, Inc. All rights reserved.
```
rmfitzpatrick: Not familiar with an official policy, but in the past I've been instructed to bump copyright years on modification/creation.

signalfx-java/src/main/java/com/signalfx/signalflow/StreamRequestException.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package com.signalfx.signalflow;
+
+import java.util.List;
+
+/**
+ * Exception thrown when the computation failed after being started.
+ *
+ * @author dgriff
```
rmfitzpatrick: Correct this?

cory-signalfx:
Fixed 'em up @rmfitzpatrick!

signalfx-java/src/main/java/com/signalfx/signalflow/ChannelMessage.java
```diff
@@ -404,11 +404,15 @@ public long getTimestampMs() {
...
-     * Message received when the computation encounters errors during its initialization.
+     * Message received when the computation encounters errors during its initialization. Because the error that is
+     * returned might have one of two sets of contents, this has both an `errors` and a `message` accessor. We'll
+     * be liberal here so the code can decide how to show this error later. Optimally, these should be different
+     * message types, but that's a lot work work.
      */
     public static class ErrorMessage extends ChannelMessage {
 
         protected ArrayList<Object> errors;
+        protected String message;
```
mpetazzoni: I think you should also capture and expose `errorCode` so you can properly differentiate between 4xx and 5xx classes of errors.

signalfx-java/src/main/java/com/signalfx/signalflow/ChannelMessage.java
```diff
@@ -404,11 +404,15 @@ public long getTimestampMs() {
     }
 
     /**
-     * Message received when the computation encounters errors during its initialization.
+     * Message received when the computation encounters errors during its initialization. Because the error that is
```
mpetazzoni: Break and new empty line before "Because the error"

signalfx-java/src/main/java/com/signalfx/signalflow/Computation.java
```diff
@@ -305,7 +305,13 @@ private void parseNext() throws ComputationAbortedException,
 
                 case ERROR_MESSAGE:
                     ErrorMessage errorMessage = (ErrorMessage) message;
-                    throw new ComputationFailedException(errorMessage.getErrors());
+                    // This is a hack based on the fact that the API can return type different error messages with
```
mpetazzoni: Use `/* ... */` for multiline comments

signalfx-java/src/main/java/com/signalfx/signalflow/Computation.java
```diff
@@ -305,7 +305,13 @@ private void parseNext() throws ComputationAbortedException,
...
+                    // the same type. We have to check attributes to know which error we're working with.
+                    if(errorMessage.message != null) {
```
mpetazzoni: Space before open paren

signalfx-java/src/main/java/com/signalfx/signalflow/StreamRequestException.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @author cwatson
+ */
+public class StreamRequestException extends RuntimeException {
+
+    private static final long serialVersionUID = 1L;
+
+    protected String message;
+
+    public StreamRequestException(String message) {
```
mpetazzoni: As discussed above I think this should also include the `errorCode`

signalfx-java/src/main/java/com/signalfx/signalflow/StreamRequestException.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (C) 2019 SignalFx, Inc. All rights reserved.
+ */
+package com.signalfx.signalflow;
+
+import java.util.List;
+
+/**
+ * Exception thrown when the computation failed after being started.
```
mpetazzoni: That seems like copy-pasta; this exception is for errors requesting the computation.

cory-signalfx: More that I am not sure what to call it. 😄

cory-signalfx:
Back your way again, @mpetazzoni. Lemme know if you have some suggestions for anything we might want to do to differentiate between the different error codes. (Note I called it `code` in the object since that's what the JSON has. Since it should only show up for errors I used `errorCode` in the exception.)

signalfx-java/src/main/java/com/signalfx/signalflow/ChannelMessage.java
```diff
@@ -405,21 +405,39 @@ public long getTimestampMs() {
...
+        protected String message;
 
         public ErrorMessage() {
             this.channelMessageType = Type.ERROR_MESSAGE;
         }
 
+        /**
+         * @return The error number, akin to an HTTP error code.
+         */
+        public int getError() { return this.error; }
```
mpetazzoni: Don't fold the method

signalfx-java/src/main/java/com/signalfx/signalflow/Computation.java
```diff
@@ -305,7 +305,15 @@ private void parseNext() throws ComputationAbortedException,
 
                 case ERROR_MESSAGE:
                     ErrorMessage errorMessage = (ErrorMessage) message;
-                    throw new ComputationFailedException(errorMessage.getErrors());
+                    /* This is a hack based on the fact that the API can return type different
+                     * error messages with the same type. We have to check attributes to know
+                     * which error we're working with.
+                     */
+                    if (errorMessage.message != null) {
```
mpetazzoni: Use the getter

signalfx-java/src/main/java/com/signalfx/signalflow/StreamRequestException.java
```diff
@@ -0,0 +1,30 @@
...
+
+    protected int errorCode;
+    protected String message;
+
+    public StreamRequestException(int errorCode, String message) {
+        super("Computation failed (" + message + ") code: " + errorCode);
+        this.message = message;
+    }
+
+    public int getErrorCode() { return this.errorCode; }
```
mpetazzoni: Don't fold the method

mpetazzoni:
You obviously meant you called it `error` in the object. That's fine.

cory-signalfx:
Fixed up and ready again. Once tests pass will call it g2g.

PR #75, opened 2019-06-04
cleanup SignalFxReporterTest

udaysagar2177:


udaysagar2177:
Had to remove test for distributed counter because it didn't make sense to include it

PR #77, opened 2019-09-05
Fix code example in README.md

timvlaer:
The example code doesn't compile. I made two changes:
* finished the builder
* Use MetricRegistry of the codahale metrics library (I think the current code still refers to yammer, since it's Metric*s*Registry)


mpetazzoni:
Thanks!

PR #78, opened 2019-09-27
Update Jetty version

artem-smotrakov:
During build, `maven-shade-plugin` copies a number of classes from `jetty` to the resulting jar file. The classes are available in `com.signalfx.shaded.jetty` package. Unfortunately, `signalfx-java` uses an old version of `jetty` which is affected by several vulnerabilities:

- CVE-2017-7658
- CVE-2017-9735
- CVE-2019-10241
- CVE-2019-10246

I have verified that the vulnerable classes are included to the resulting jar, and can be accessed by an application if only `signalfx-java` is added as a dependency (`jetty` doesn't have to added explicitly). 

Strictly speaking, it seems to be extremely difficult to exploit the issues above if an application just uses `signalfx-java`. I am not putting all the details here, but please let me know if you are interested. However, it might be better to update jetty to be on the safe side. Furthermore, dependency scanners like [Vulas](https://github.com/SAP/vulnerability-assessment-tool) report the issues above if `signalfx-java` is used by an application. Updating `jetty` version may make user's life a bit easier.

The patch updates `jetty` to the latest version in the `9.x` branch.

I am also wondering if it is necessary to relocate jetty classes with `maven-shade-plugin`. I am not an expert in `signalfx-java` but I see that `jetty-server` is added as a test dependency. Looks like it's used only for testing but not required by applications which use `signalfx-java`.

Please let me know if `jetty` may be removed from the configuration of `maven-shade-plugin`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,7 +27,7 @@
     <os72.protobuf-shaded.plugin-version>${protobuf.version}.1</os72.protobuf-shaded.plugin-version>
     <jackson.version>2.9.9</jackson.version>
     <jackson.databind.version>2.9.9.3</jackson.databind.version>
-    <jetty.version>9.2.26.v20180806</jetty.version>
+    <jetty.version>9.2.28.v20190418</jetty.version>
   </properties>
 
   <url>http://www.signalfx.com</url>


mpetazzoni:
Unfortunately as long as we need to support Java 7 we cannot move past Jetty 9.2. See https://github.com/signalfx/signalfx-java/commit/6c708b3285a286cd037853c90cba9e269157ac5f for more details.

We might be able to move to 9.2.28.v20190418 if you can confirm that version includes the security fixes for the CVEs you have listed above.

artem-smotrakov:
@mpetazzoni No problems, `9.2.28.v20190418` includes the fixes for the issues above. I've updated the patch.

PR #82, opened 2020-01-10
Drop support for Java 6 and 7

keitwb:
- Also upgrade Jetty now that we are on 8+


keitwb:
> LGTM. When we release we should make a major version change (might be time to go to 1.x !)

Yes, will go to 1.0.0 as this will be breaking for anybody using Java 6/7.

PR #91, opened 2020-07-09
Bug fix for sending metric with connection pool

wxy325:
For the case when using `PoolingHttpClientConnectionManager`, e.g.
```java
SignalFxReceiverEndpoint endpoint =
      new SignalFxEndpoint(ingestUrl.getProtocol(), ingestUrl.getHost(), ingestUrl.getPort());

PoolingHttpClientConnectionManager m = new PoolingHttpClientConnectionManager(); 
HttpDataPointProtobufReceiverFactory f = new HttpDataPointProtobufReceiverFactory(endpoint).setVersion(2);
f.setHttpClientConnectionManager(m); // We pass a HTTP connection pool here (the default connection pool for apache http components) in order to support HTTP keep-alive

AggregateMetricSender mf = new AggregateMetricSender("SendMetrics",
      f,
      new HttpEventProtobufReceiverFactory(endpoint),
      new StaticAuthToken("<Org Token>"),
      Collections.<OnSendErrorHandler>singleton(new OnSendErrorHandler() {
        @Override
        public void handleError(MetricError metricError) {
          System.out.println("Unable to POST metrics to SignalFx: " + metricError.getMessage() + " with errorType:" + metricError.getMetricErrorType());
        }
      }));
     

for (true) {
    // dp = get a batch of metric data somewhere
    AggregateMetricSender.Session session = mf.createSession();
    session.setDatapoint(dp);
    try {
        session.close();
    } catch (IOException ex) {
        System.out.println("Problem closing AggregateMetricSender.Session object");
        throw ex;
    }
}
```
Calling `resp.close()` directly will terminate the reuse-able TCP connection, while calling `resp.getEntity().getContent().close()` before it will release the TCP connection back to pool and reuse it in the future (for HTTP keep-alive) (Probably a bug for apache HTTP components)

Network packets screenshoots (captured by Wireshark) for sending 3 batches of metrics by above sample follow:

Before this MR, we can observe 3 TCP connections, each of which sends one batch of metric
sfx_metric_1
sfx_metric_2
sfx_metric_3

After this MR, we can observe 1 TCP connection, which sends 3 batches of metric
after_fix

Feel free to ping @ xwu in Splunk's slack workspace if there are any question, thanks!

### DIFF
--- a/signalfx-java/src/main/java/com/signalfx/metrics/connection/AbstractHttpDataPointProtobufReceiverConnection.java
+++ b/signalfx-java/src/main/java/com/signalfx/metrics/connection/AbstractHttpDataPointProtobufReceiverConnection.java
@@ -55,6 +55,10 @@ public abstract class AbstractHttpDataPointProtobufReceiverConnection extends Ab
                 }
             } finally {
                 if (resp != null) {
+                    HttpEntity entity = resp.getEntity();
+                    if (entity != null) {
+                        entity.getContent().close();
+                    }
                     resp.close();
                 }
             }
@@ -102,6 +106,10 @@ public abstract class AbstractHttpDataPointProtobufReceiverConnection extends Ab
                 }
             } finally {
                 if (resp != null) {
+                    HttpEntity entity = resp.getEntity();
+                    if (entity != null) {
+                        entity.getContent().close();
+                    }
                     resp.close();
                 }
             }


signalfx-java/src/main/java/com/signalfx/metrics/connection/AbstractHttpDataPointProtobufReceiverConnection.java
```diff
@@ -55,6 +55,7 @@ public void addDataPoints(String auth, List<SignalFxProtocolBuffers.DataPoint> d
                 }
             } finally {
                 if (resp != null) {
+                    resp.getEntity().getContent().close();
```
keitwb, line 58: Are you fairly certain that `getEntity()` will always return non-null?  According to [docs](https://hc.apache.org/httpcomponents-core-ga/httpcore/apidocs/org/apache/http/HttpResponse.html#getEntity()) it can return null (looks like `getContent()` is safe though). E.g. maybe if there is a proxy that returns no body for whatever reason (our own ingest server/gateway would always return a body but no guarantees).

wxy325: @keitwb Thanks for checking it! I did not do null check previously under the assumption that the ingest endpoint will always return body, and the null check has been added in the latest commit

PR #114, opened 2021-04-30
bump commons-io version from 2.6 to 2.7 (mitigate CVE-2021-29425)

breedx-splk:
Dependabot flagged this, so let's bump the version.


breedx-splk:
OMG derp I did not see that dependabot also created a PR.

PR #129, opened 2021-06-25
Clean up pom.xml files

mateuszrzeszutek:
* Remove a few unused libs
* Categorize deps by usage (shaded, compile, test) in the parent pom.xml


laurit (approved):

signalfx-codahale/pom.xml
```diff
@@ -61,20 +62,18 @@
       <artifactId>metrics-core</artifactId>
     </dependency>
 
+    <!-- test -->
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-simple</artifactId>
-      <scope>test</scope>
```
jkwatson, line 67: why remove the scope here?

jkwatson: oh, does it get it inherited from the main pom?

mateuszrzeszutek: Yes, I removed all inherited test scopes

PR #130, opened 2021-06-25
Add GitLab pipeline

mateuszrzeszutek:


jkwatson (approved):
I will be doing the same for Android soon

.gitlab-ci.yml
```diff
@@ -0,0 +1,19 @@
...
+  name: "openjdk:11.0.11-9-jdk"
+
+stages:
+  - build
+  - release
+
+build:
+  stage: build
+  script:
+    - ./mvnw clean install -DskipTests
```
laurit, line 11: github uses `./mvnw clean package -DskipTests` and `./mvnw verify` any reason why it is different here? Before https://github.com/signalfx/signalfx-java/pull/119 github used the same commands as here.

laurit (approved):

.gitlab-ci.yml
mateuszrzeszutek: No reason -- this one can use `package` too; I just typed those commands from the memory and not copied them 😅

laurit: Actually github uses `./mvnw test` instead of `./mvnw verify`, meant to write it in the first comment but messed up :)

mateuszrzeszutek: Okay, now they're the same 😛

jkwatson: I *think* `verify` is the preferred maven way to do things now, FYI.

mateuszrzeszutek: OK, let's use verify everywhere then 👍

PR #131, opened 2021-06-25
Deploy snapshots and releases

mateuszrzeszutek:


deploy-release.sh
```diff
@@ -0,0 +1,42 @@
...
+        <activeByDefault>false</activeByDefault>
+      </activation>
+      <properties>
+        <gpg.passphrase>${GPG_PASSWORD}</gpg.passphrase>
+      </properties>
+    </profile>
+  </profiles>
+</settings>
+EOF
+trap "rm release-settings.xml" EXIT
```
breedx-splk, line 40: Fancy!  🐱

jkwatson (approved):

breedx-splk (approved):
Seems good to me. I like the use of `trap` here.

deploy-release.sh
mateuszrzeszutek: Turns out you can't set >1 trap for a single signal in bash, so I had to change that a little...

mateuszrzeszutek: Well I ended up using `trap` after all 😅

pom.xml
```diff
@@ -99,6 +87,19 @@
     </profile>
   </profiles>
 
+  <distributionManagement>
+    <snapshotRepository>
+      <id>ossrh</id>
```
iNikem, line 92: What is `rh`?

iNikem (approved):

pom.xml
mateuszrzeszutek: `OSS Repository Hosting` - I copied it from Sonatype docs basically 😅

PR #159, opened 2021-08-18
Fix branch name in GL CI

mateuszrzeszutek:


.gitlab-ci.yml
```diff
@@ -20,7 +20,7 @@ build:
 sast-scan:
   stage: verify
   rules:
-    - if: '$CI_COMMIT_REF_NAME == "main"'
+    - if: '$CI_COMMIT_REF_NAME == "master"'
```
jkwatson, line 23: wait...why is this going *to* master from main? Shouldn't we be using main everywhere?

mateuszrzeszutek: We haven't done the master->main migration in all repos, only those that are going to be supported long term I guess?

mateuszrzeszutek: Although we're probably going to support this repo for a while; it'd make sense to do the migration.

jkwatson (approved):

breedx-splk:
Tracked it with #160 .

PR #206, opened 2022-03-15
Change name of primary branch to main

### LINKED ISSUE #160: Change primary branch name from `master` to `main`
This is newer, less troubling naming convention and we should migrate this repo.

breedx-splk:
This will resolve #160. 

Once this is merged, we will need to change the setting (which I do not have access to).


jkwatson (approved):
Thanks!

Issue #189, opened 2022-01-14
Provide unshaded version or relocate additional dependencies


pkwarren:
When looking at the latest signalfx-java 1.0.14 artifact, I show these dependencies as shaded without relocation (causing duplicate classes on the classpath). If an unshaded version is made available, teams could ensure consistent versions across an application. Alternatively, all of these dependencies should be relocated to minimize conflicts on the classpath.

```
com/google/errorprone/**/*.class
com/google/j2objc/**/*.class
google/protobuf/**/*.proto
javax/annotation/**/*.class
mozilla/public-suffix-list.txt (I believe this comes with Guava?)
org/checkerframework/**/*.class
```

We encountered this because of running security checks which found vulnerabilities in shaded Jackson and other libraries in an older version. If this artifact was available unshaded, we'd have been able to manage the versions of dependencies across all components.

helloLiRong:
any update??

We are facing the same issue. I would suggest excluding those packages when relocating. Only relocate the classpath that you have to.

anthochristen:
Facing the same. Most of them are transitive dependencies from `guava` which causes an issue if we also have guava along with singal-fx on the classpath.

mateuszrzeszutek:
@pkwarren @helloLiRong @anthochristen 
Sorry for the long wait, it seems we haven't noticed this issue before. All the requested changes have been implemented, all dependencies are now being shaded - we'll release the 1.0.18 version with all these fixes shortly.

PR #227, opened 2022-04-16
Exclude google *.proto files

luke-sterkowicz:
They lead to conflicts and cause problems for protobuf-maven-plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -294,6 +294,14 @@
         <configuration>
           <createSourcesJar>true</createSourcesJar>
           <shadeSourcesContent>true</shadeSourcesContent>
+          <filters>
+            <filter>
+              <artifact>*:*</artifact>
+              <excludes>
+                <exclude>google/protobuf/**/*.proto</exclude>
+              </excludes>
+            </filter>
+          </filters>
           <artifactSet>
             <excludes>
               <exclude>org.slf4j:*</exclude>


breedx-splk (commented):
Any more specific detail about how this breaks the plugin?

I'm not aware of any reason why we would include these in the archive....and consumers that need them can source them on their own, right? Are these a transitive dependency? I didn't see them explicitly included anywhere.

Do we consider this a breaking change (because users could be relying on them being there)?  I think this is good, I just want to make sure we've thought about it.

luke-sterkowicz:
Sure, protobuf-maven-plugin extracts *.proto files from all dependencies to allow referencing them from your project's *.proto during code generation. With google definitions included in signalfx-java, they will be duplicated with the right (google) ones (but providing shadowed java packages), and  may be referenced instead. If that happens (it actually did when I tested it), protobuf-maven-plugin generates invalid sources with mixed packages, which don't compile.

Why they appear in signalfx-java? It's because maven-shade-plugin cares only about *.class and copies everything else straight by default.

I don't consider it a breaking change. If anybody makes use of these files, they shouldn't (should take from the source - i.e. protobuf-java or some libs which provide already shadowed versions, such as protobuf-java-shaded-xxx which is used in this very project).
It was actually requested to be cleaned up, see https://github.com/signalfx/signalfx-java/issues/189, although I'm not knowledgeable about other files mentioned and whether they're used anyhow - just want to fix the case I'm aware of.

mateuszrzeszutek (approved):

mateuszrzeszutek:
> It was actually requested to be cleaned up, see #189, although I'm not knowledgeable about other files mentioned and whether they're used anyhow - just want to fix the case I'm aware of.

I kinda missed that issue (didn't subscribe to all notifications...), most of these are already fixed; I'll fix the rest in another PR.

Thanks!

PR #240, opened 2022-06-30
Force apache commons-codec to 1.15

breedx-splk:
Mitigates https://security.snyk.io/vuln/SNYK-JAVA-COMMONSCODEC-561518

It sucks to do it this way, but it looks like the v4.x release of httpclient hasn't yet been updated (see the bottom comment from 6 months ago) https://issues.apache.org/jira/browse/HTTPCLIENT-2072

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -156,6 +156,12 @@
         <version>2.11.0</version>
       </dependency>
       <dependency>
+        <!-- Temp work-around until httpclient upgrades its codec version -->
+        <groupId>commons-codec</groupId>
+        <artifactId>commons-codec</artifactId>
+        <version>1.15</version>
+      </dependency>
+      <dependency>
         <groupId>org.apache.httpcomponents</groupId>
         <artifactId>httpclient</artifactId>
         <version>4.5.13</version>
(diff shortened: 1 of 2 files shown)


mateuszrzeszutek (approved):

signalfx-java/pom.xml
```diff
@@ -80,6 +80,12 @@
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-annotations</artifactId>
     </dependency>
+    <dependency>
+      <!-- Temp work-around until httpclient upgrades its codec version -->
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+      <version>1.15</version>
```
laurit, line 87: you don't need to add the version number here, having it in the parent pom is enough. None of the other dependencies in this file add a version number. Another option is to not add this dependency here at all, having it in parent is sufficient.

pom.xml
```diff
@@ -155,6 +155,12 @@
         <artifactId>commons-io</artifactId>
         <version>2.11.0</version>
       </dependency>
+      <dependency>
+        <!-- Temp work-around until httpclient upgrades its codec version -->
```
laurit, line 159: this comment is a bit misleading. last release from 4.5.x series of http client was in 2020

PR #241, opened 2022-07-04
v1.0.20 pre-release changes

laurit:


signalfx-java/pom.xml
```diff
@@ -80,12 +80,6 @@
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-annotations</artifactId>
     </dependency>
-    <dependency>
```
laurit, line 83: this is already listed in dependency management in parent pom

mateuszrzeszutek (approved):

PR #243, opened 2022-07-06
Yield message on END_OF_CHANNEL

jtrzebiatowski:
This change makes it possible to process END_OF_CHANNEL channel messages when iterating over computation.

```
for (ChannelMessage message : computation) {
            switch (message.getType()) {
                case DATA_MESSAGE, EVENT_MESSAGE, METADATA_MESSAGE, JOB_PROGRESS, JOB_START, STREAM_START, EXPIRED_TSID_MESSAGE, INFO_MESSAGE, ERROR_MESSAGE:
                    break;
                case CHANNEL_ABORT:
                    return;
                case END_OF_CHANNEL:
                    // this is now possible.
                    return;
            }
        }
```

### DIFF
--- a/signalfx-java/src/main/java/com/signalfx/signalflow/Computation.java
+++ b/signalfx-java/src/main/java/com/signalfx/signalflow/Computation.java
@@ -239,6 +239,7 @@ public class Computation implements Iterable<ChannelMessage>, Iterator<ChannelMe
 
                 case END_OF_CHANNEL:
                     state = State.STATE_COMPLETED;
+                    nextMessage = message;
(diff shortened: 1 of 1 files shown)


jtrzebiatowski:
Hello @laurit ,
Could you please review this change? Thanks in advance.

I didn't see any tests for this class and scenario.

jtrzebiatowski:
I now see this actually not needed because the iterator ends when END_OF_CHANNEL is emitted.

I'm closing the PR. If someone finds this useful feel free to copy it.

PR #249, opened 2022-07-26
Add support to emit delta histogram count buckets for Timer and DistributionSummary

bogdandrutu:
Current PR adds a new configuration "publishDeltaHistogram" which has higher priority than "publishCumulativeHistogram" when configure both.

The current logic is trivial, and uses the cumulative implementation and calculates delta every time the snapshot is taken, as a diff between the current value and the last snapshot.

Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>

### DIFF
--- a/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/DeltaHistogramSnapshot.java
+++ b/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/DeltaHistogramSnapshot.java
@@ -20,23 +20,14 @@ import io.micrometer.core.instrument.distribution.CountAtBucket;
 import io.micrometer.core.instrument.distribution.HistogramSnapshot;
 
 final class DeltaHistogramSnapshot {
-    // It may get called from different threads, so use volatile to ensure updates are visible.
-    // Not null only if producing delta.
-    private volatile HistogramSnapshot lastSnapshot;
+    private HistogramSnapshot lastSnapshot;
 
-    DeltaHistogramSnapshot(boolean isDelta) {
-        if (isDelta) {
-            lastSnapshot = HistogramSnapshot.empty(0, 0, 0);
-        } else {
-            lastSnapshot = null;
-        }
+    DeltaHistogramSnapshot() {
+        lastSnapshot = HistogramSnapshot.empty(0, 0, 0);
     }
 
     // TODO: Determine if we need to synchronize, in case multiple calls in parallel.
     HistogramSnapshot calculateSnapshot(HistogramSnapshot currentSnapshot) {
-        if (lastSnapshot == null) {
-            return currentSnapshot;
-        }
         HistogramSnapshot deltaSnapshot = new HistogramSnapshot(
                 currentSnapshot.count() - lastSnapshot.count(),
                 currentSnapshot.total() - lastSnapshot.total(),
--- a/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/DeltaHistogramSnapshot.java
+++ b/micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/DeltaHistogramSnapshot.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright Splunk Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.micrometer.signalfx;
+
+import io.micrometer.core.instrument.distribution.CountAtBucket;
+import io.micrometer.core.instrument.distribution.HistogramSnapshot;
+
+final class DeltaHistogramSnapshot {
+    // It may get called from different threads, so use volatile to ensure updates are visible.
+    // Not null only if producing delta.
+    private volatile HistogramSnapshot lastSnapshot;
+
+    DeltaHistogramSnapshot(boolean isDelta) {
+        if (isDelta) {
+            lastSnapshot = HistogramSnapshot.empty(0, 0, 0);
+        } else {
+            lastSnapshot = null;
+        }
+    }
+
+    // TODO: Determine if we need to synchronize, in case multiple calls in parallel.
+    HistogramSnapshot calculateSnapshot(HistogramSnapshot currentSnapshot) {
+        if (lastSnapshot == null) {
+            return currentSnapshot;
+        }
+        HistogramSnapshot deltaSnapshot = new HistogramSnapshot(
+                currentSnapshot.count() - lastSnapshot.count(),
+                currentSnapshot.total() - lastSnapshot.total(),
+                currentSnapshot.max(),  // Max cannot be calculated as delta, keep the current.
+                null,  // No percentile values
+                deltaHistogramCounts(currentSnapshot),
+                currentSnapshot::outputSummary);
+        lastSnapshot = currentSnapshot;
+        return deltaSnapshot;
+    }
+
+    private CountAtBucket[] deltaHistogramCounts(HistogramSnapshot currentSnapshot) {
+        CountAtBucket[] currentHistogramCounts = currentSnapshot.histogramCounts();
+        CountAtBucket[] lastHistogramCounts = lastSnapshot.histogramCounts();
+        if (lastHistogramCounts == null || lastHistogramCounts.length == 0) {
+            return currentHistogramCounts;
+        }
+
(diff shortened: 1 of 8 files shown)


micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxConfig.java
```diff
@@ -64,14 +64,25 @@ default String accessToken() {
     }
 
     /**
-     * @return {@code true} if the SignalFx registry should emit cumulative histogram
-     * buckets.
+     * If both "publishCumulativeHistogram" and "publishDeltaHistogram" are set, then delta will be used.
```
breedx-splk, line 67: The both thing is confusing...but I can live with it. Preference would be to have a single setting (something like `histogramBucketingStrategy`) with an enumeration of values like `delta` and `cumulative`.

micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalfxTimer.java
```diff
@@ -62,12 +63,15 @@ final class SignalfxTimer extends AbstractTimer {
...
+    private final DeltaHistogramSnapshot deltaHistogramSnapshot;
+
     SignalfxTimer(Id id, Clock clock, DistributionStatisticConfig distributionStatisticConfig,
-            PauseDetector pauseDetector, TimeUnit baseTimeUnit, long stepMillis) {
+            PauseDetector pauseDetector, TimeUnit baseTimeUnit, long stepMillis, boolean isDelta) {
         super(id, clock, CumulativeHistogramConfigUtil.updateConfig(distributionStatisticConfig), pauseDetector,
                 baseTimeUnit, false);
         countTotal = new StepTuple2<>(clock, stepMillis, 0L, 0L, count::sumThenReset, total::sumThenReset);
         max = new TimeWindowMax(clock, distributionStatisticConfig);
+        deltaHistogramSnapshot = new DeltaHistogramSnapshot(isDelta);
```
breedx-splk, line 74: Better DI practice is to pass in the `deltaHistogramSnapshot` instance rather than doing the creation here.

micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
```diff
@@ -165,23 +168,23 @@ protected void publish() {
...
     @Override
     protected DistributionSummary newDistributionSummary(Meter.Id id,
             DistributionStatisticConfig distributionStatisticConfig, double scale) {
-        if (!publishCumulativeHistogram) {
+        if (!publishCumulativeHistogram && !publishDeltaHistogram) {
             return super.newDistributionSummary(id, distributionStatisticConfig, scale);
         }
         DistributionSummary summary = new SignalfxDistributionSummary(id, clock, distributionStatisticConfig, scale,
-                config.step().toMillis());
+                config.step().toMillis(), this.publishDeltaHistogram);
```
breedx-splk, line 187:
```suggestion
                config.step().toMillis(), publishDeltaHistogram);
```

micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalfxDistributionSummary.java
```diff
@@ -61,11 +62,14 @@ final class SignalfxDistributionSummary extends AbstractDistributionSummary {
...
+    private final DeltaHistogramSnapshot deltaHistogramSnapshot;
+
     SignalfxDistributionSummary(Id id, Clock clock, DistributionStatisticConfig distributionStatisticConfig,
-            double scale, long stepMillis) {
+            double scale, long stepMillis, boolean isDelta) {
         super(id, clock, CumulativeHistogramConfigUtil.updateConfig(distributionStatisticConfig), scale, false);
         this.countTotal = new StepTuple2<>(clock, stepMillis, 0L, 0.0, count::sumThenReset, total::sumThenReset);
-        max = new TimeWindowMax(clock, distributionStatisticConfig);
+        this.max = new TimeWindowMax(clock, distributionStatisticConfig);
+        this.deltaHistogramSnapshot = new DeltaHistogramSnapshot(isDelta);
```
breedx-splk, line 72: Same comment about passing in the actual dependency instead of creating an instance here in the constructor.

breedx-splk (changes requested):
Had a few small requests. I am also kinda concerned that we are adding custom code to this directory, which is effectively a fork. The readme says that it's a one time fork that will not be updated, so this PR goes against that thought.  Maybe @mateuszrzeszutek has some thoughts as well...

micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxConfig.java
bogdandrutu: Correct, but the previous one already exists, so will be a breaking change.

micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/DeltaHistogramSnapshot.java
```diff
@@ -0,0 +1,66 @@
...
+
+import io.micrometer.core.instrument.distribution.CountAtBucket;
+import io.micrometer.core.instrument.distribution.HistogramSnapshot;
+
+final class DeltaHistogramSnapshot {
+    // It may get called from different threads, so use volatile to ensure updates are visible.
+    // Not null only if producing delta.
+    private volatile HistogramSnapshot lastSnapshot;
+
+    DeltaHistogramSnapshot(boolean isDelta) {
```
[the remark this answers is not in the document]
bogdandrutu: I went ahead an move the logic of producing "delta" or not in the parent class, so this is initialize only when deltas are requested.

micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalfxTimer.java
bogdandrutu: I agree, but the whole code in micrometer is written this way, look at the `DistributionStatisticConfig` which passes the config and the class receiving the configuration creates the necessary objects. Will be an anti pattern when trying to upstream the code.

micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalfxDistributionSummary.java
bogdandrutu: See previous response. I will resolve this comment and continue the discussion there, if we decide something will make sure will apply here as well.

micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalFxMeterRegistry.java
bogdandrutu: Done, in both cases.

micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/SignalfxTimer.java
breedx-splk: Yeah ok. I guess we'll favor consistency then.

micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/DeltaHistogramSnapshot.java
```diff
@@ -0,0 +1,66 @@
...
+
+        CountAtBucket[] retHistogramCounts = new CountAtBucket[currentHistogramCounts.length];
+        for (int i = 0; i < currentHistogramCounts.length; i++) {
+            retHistogramCounts[i] = new CountAtBucket(
+                    currentHistogramCounts[i].bucket(),
+                    currentHistogramCounts[i].count() - lastHistogramCounts[i].count());
+        }
+        return retHistogramCounts;
+    }
+}
```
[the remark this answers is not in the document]
bogdandrutu: Done.

bogdandrutu:
> The readme says that it's a one time fork that will not be updated, so this PR goes against that thought. Maybe @mateuszrzeszutek has some thoughts as well...

We discussed internally, and decision was to make this contribution initially then upstream the change (which takes way longer than our deadline request).

micrometer-registry-signalfx/src/main/java/io/micrometer/signalfx/DeltaHistogramSnapshot.java
```diff
@@ -0,0 +1,66 @@
...
+
+    DeltaHistogramSnapshot(boolean isDelta) {
+        if (isDelta) {
+            lastSnapshot = HistogramSnapshot.empty(0, 0, 0);
+        } else {
+            lastSnapshot = null;
+        }
+    }
+
+    // TODO: Determine if we need to synchronize, in case multiple calls in parallel.
```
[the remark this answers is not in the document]
bogdandrutu: Looking into different other implementation in micrometer world, see https://github.com/micrometer-metrics/micrometer/blob/c7262538893a6c1fb9581775082df978f8a1c722/micrometer-core/src/main/java/io/micrometer/core/instrument/internal/CumulativeHistogramLongTaskTimer.java#L34 it seems that there is no need for any threading guarantees. Will follow the same.

breedx-splk (approved):

PR #250, opened 2022-07-27
Prepare for the 1.0.21 release

mateuszrzeszutek:


update_version.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
mateuszrzeszutek, line 1: There's no python2 on the latest MacOS version anymore

laurit (approved):

PR #255, opened 2022-07-29
Make SignalFlowClient implement AutoCloseable

einarpehrson:
To enable the use of try-with-resources.

### DIFF
--- a/signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
+++ b/signalfx-java/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
@@ -18,7 +18,7 @@ import com.google.common.collect.ImmutableMap;
  *
  * @author dgriff
  */
-public class SignalFlowClient {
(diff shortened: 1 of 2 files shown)


breedx-splk (changes requested):
If your intent is to use try-with-resources, I think this should be using `AutoCloseable` right? Also please add the `@Override` annotation and add a test that demonstrates try-with-resources.

einarpehrson:
> If your intent is to use try-with-resources, I think this should be using `AutoCloseable` right?
@breedx-splk yes, you are right. Fixed.

> Also please add the `@Override` annotation and add a test that demonstrates try-with-resources.
@breedx-splk Done.

breedx-splk (approved):

einarpehrson:
@breedx-splk would you mind merging this pull request? :)

mateuszrzeszutek:
Merged. Thanks @einarpehrson for your contribution!

PR #257, opened 2022-08-03
Try improving the ref name

breedx-splk:
It seems to work for other similar repos, if we can use `latest` that would be much preferred.


mateuszrzeszutek (approved):

breedx-splk:
Well that failed spectacularly....

image

Not sure why it works for other projects. I guess I'll change it back until prodsec can get their names fixed.

PR #268, opened 2022-08-25
Make retryable exceptions configurable

Excaleo:
### Context

Recently we've been seeing read socket timeouts from SignalFx on our end. We've increased the timeout of the connection to 15 seconds, however we are still seeing `SocketTimeoutException` which leads to data loss. To circumvent this and prevent data loss, we would like to retry on this exception. Therefore, I have made the list of exceptions to retry on configurable, with a default list which replicates the previous behaviour

### Changes implemented

- Add configurable exceptions to retry logic via constructor overloading for the datapoints API

### DIFF
--- a/signalfx-java/src/main/java/com/signalfx/connection/RetryDefaults.java
+++ b/signalfx-java/src/main/java/com/signalfx/connection/RetryDefaults.java
@@ -11,7 +11,7 @@ public final class RetryDefaults {
     private RetryDefaults() {
     }
 
-    public static final Integer DEFAULT_MAX_RETRIES = 3;
+    public static final int DEFAULT_MAX_RETRIES = 3;
     public static final List<Class<? extends IOException>> DEFAULT_NON_RETRYABLE_EXCEPTIONS = Arrays.asList(
             InterruptedIOException.class,
             UnknownHostException.class,
--- a/signalfx-java/src/main/java/com/signalfx/connection/RetryDefaults.java
+++ b/signalfx-java/src/main/java/com/signalfx/connection/RetryDefaults.java
@@ -5,6 +5,7 @@ import java.io.InterruptedIOException;
 import java.net.ConnectException;
 import java.net.UnknownHostException;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 
 public final class RetryDefaults {
@@ -12,8 +13,8 @@ public final class RetryDefaults {
     }
 
     public static final int DEFAULT_MAX_RETRIES = 3;
-    public static final List<Class<? extends IOException>> DEFAULT_NON_RETRYABLE_EXCEPTIONS = Arrays.asList(
+    public static final List<Class<? extends IOException>> DEFAULT_NON_RETRYABLE_EXCEPTIONS = Collections.unmodifiableList(Arrays.asList(
             InterruptedIOException.class,
             UnknownHostException.class,
-            ConnectException.class);
+            ConnectException.class));
 }
--- a/signalfx-java/src/main/java/com/signalfx/connection/RetryDefaults.java
+++ b/signalfx-java/src/main/java/com/signalfx/connection/RetryDefaults.java
@@ -1,0 +1,19 @@
+package com.signalfx.connection;
+
+import java.io.IOException;
+import java.io.InterruptedIOException;
+import java.net.ConnectException;
+import java.net.UnknownHostException;
+import java.util.Arrays;
+import java.util.List;
+
+public final class RetryDefaults {
+    private RetryDefaults() {
+    }
+
+    public static final Integer DEFAULT_MAX_RETRIES = 3;
+    public static final List<Class<? extends IOException>> DEFAULT_NON_RETRYABLE_EXCEPTIONS = Arrays.asList(
+            InterruptedIOException.class,
+            UnknownHostException.class,
+            ConnectException.class);
+}
--- a/signalfx-java/src/main/java/com/signalfx/metrics/connection/HttpDataPointProtobufReceiverFactory.java
+++ b/signalfx-java/src/main/java/com/signalfx/metrics/connection/HttpDataPointProtobufReceiverFactory.java
@@ -7,6 +7,7 @@ import com.signalfx.endpoint.SignalFxReceiverEndpoint;
 import com.signalfx.metrics.SignalFxMetricsException;
 
 import java.io.IOException;
+import java.util.Collections;
 import java.util.List;
 
 import static com.signalfx.connection.RetryDefaults.DEFAULT_MAX_RETRIES;
@@ -49,7 +50,7 @@ public class HttpDataPointProtobufReceiverFactory implements DataPointReceiverFa
     }
 
     public HttpDataPointProtobufReceiverFactory setNonRetryableExceptions(List<Class<? extends IOException>> clazzes) {
-        this.nonRetryableExceptions = clazzes;
+        this.nonRetryableExceptions = Collections.unmodifiableList(clazzes);
         return this;
     }
 
--- a/signalfx-java/src/main/java/com/signalfx/metrics/connection/HttpDataPointProtobufReceiverFactory.java
+++ b/signalfx-java/src/main/java/com/signalfx/metrics/connection/HttpDataPointProtobufReceiverFactory.java
@@ -6,10 +6,21 @@ import com.google.common.base.MoreObjects;
 import com.signalfx.endpoint.SignalFxReceiverEndpoint;
 import com.signalfx.metrics.SignalFxMetricsException;
 
+import java.io.IOException;
+import java.io.InterruptedIOException;
+import java.net.ConnectException;
+import java.net.UnknownHostException;
+import java.util.Arrays;
+import java.util.List;
+
 public class HttpDataPointProtobufReceiverFactory implements DataPointReceiverFactory {
     public static final int DEFAULT_TIMEOUT_MS = 2000;
     public static final int DEFAULT_VERSION = 2;
     public static final int DEFAULT_MAX_RETRIES = 3;
+    public static final List<Class<? extends IOException>> DEFAULT_NON_RETRYABLE_EXCEPTIONS = Arrays.asList(
+            InterruptedIOException.class,
+            UnknownHostException.class,
+            ConnectException.class);
 
     private final SignalFxReceiverEndpoint endpoint;
     private HttpClientConnectionManager httpClientConnectionManager;
@@ -17,6 +28,7 @@ public class HttpDataPointProtobufReceiverFactory implements DataPointReceiverFa
     private int timeoutMs = DEFAULT_TIMEOUT_MS;
     private int version = DEFAULT_VERSION;
     private int maxRetries = DEFAULT_MAX_RETRIES;
+    private List<Class<? extends IOException>> nonRetryableExceptions = DEFAULT_NON_RETRYABLE_EXCEPTIONS;
 
     public HttpDataPointProtobufReceiverFactory(SignalFxReceiverEndpoint endpoint) {
         this.endpoint = endpoint;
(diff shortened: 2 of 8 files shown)


irisgve:
Hello! Any chance this can be reviewed? Thanks!

signalfx-java/src/test/java/com/signalfx/metrics/connection/HttpDataPointProtobufReceiverConnectionTest.java
```diff
@@ -95,6 +97,63 @@ public void shouldSendDataEvenIfServerClosedConnection() throws Exception {
...
+              .setNonRetryableExceptions(Collections.emptyList())
+              .createDataPointReceiver();
+      try {
+        dpr.addDataPoints(AUTH_TOKEN, Collections.singletonList(
+                SignalFxProtocolBuffers.DataPoint.newBuilder().setSource("source").build()));
+      } catch (Exception ignored2) {
+      }
+    }
+
+    Assert.assertEquals(2, handler.requests);
```
breedx-splk, line 126: static import please.

signalfx-java/src/test/java/com/signalfx/metrics/connection/HttpDataPointProtobufReceiverConnectionTest.java
```diff
@@ -95,6 +97,63 @@ public void shouldSendDataEvenIfServerClosedConnection() throws Exception {
     }
   }
 
+  @Test
+  public void shouldRetryOnSocketTimeout() throws Exception {
+    final CountingTimeoutHandler handler = new CountingTimeoutHandler();
+    final int timeout = 1000;
```
breedx-splk, line 103: inline `timeout`

signalfx-java/src/test/java/com/signalfx/metrics/connection/HttpDataPointProtobufReceiverConnectionTest.java
```diff
@@ -95,6 +97,63 @@ public void shouldSendDataEvenIfServerClosedConnection() throws Exception {
...
+              .setMaxRetries(1)
+              .createDataPointReceiver();
+      try {
+        dpr.addDataPoints(AUTH_TOKEN, Collections.singletonList(
+                SignalFxProtocolBuffers.DataPoint.newBuilder().setSource("source").build()));
+      } catch (Exception ignored2) {
+      }
+    }
+
+    Assert.assertEquals(1, handler.requests);
```
breedx-splk, line 154: static import please.

signalfx-java/src/test/java/com/signalfx/metrics/connection/HttpDataPointProtobufReceiverConnectionTest.java
```diff
@@ -193,6 +252,20 @@ public void handle(String target, Request baseRequest, HttpServletRequest reques
...
 
+  private static class CountingTimeoutHandler extends AbstractHandler {
+    private int requests = 0;
+
+    @Override
+    public void handle(String s, Request request, HttpServletRequest httpServletRequest, HttpServletResponse httpServletResponse) throws IOException{
+      requests++;
+
+      try {
+        Thread.sleep(2000);
```
breedx-splk, line 263: I'm always bummed to see tests getting slower...and in my experience tests with timeouts are often flappy/flaky when run in cloud CI/CD. Can we figure out a way to pass a `CountdownLatch` or something that the test/caller can trigger so that we don't have to actually wait?

breedx-splk (changes requested):
Thanks you for the submission. I had a few requests, but this all does seem pretty reasonable.

signalfx-java/src/test/java/com/signalfx/metrics/connection/HttpDataPointProtobufReceiverConnectionTest.java
Excaleo: I know that feel 🫂 , I saw it being used already so chucked it in. I've modified it to a hybrid approach, a latched handler that also has a timeout. If the latch is triggered manually then it will exit before, for convenience, I'm providing a timeout value on the latch for my testing purposes.

Let me know if this is good enough, or if you'd prefer I trigger the latch manually.

Excaleo:
Unrelated to the PR, but has anyone run into issues unit testing locally for `HttpDataPointProtobufReceiverConnectionTest` ? I'm seeing some timeouts for Jetty server in all the test cases, but the tests run fine in a containerized environment. 🤔

signalfx-java/src/main/java/com/signalfx/connection/RetryDefaults.java
```diff
@@ -0,0 +1,19 @@
...
+import java.net.ConnectException;
+import java.net.UnknownHostException;
+import java.util.Arrays;
+import java.util.List;
+
+public final class RetryDefaults {
+    private RetryDefaults() {
+    }
+
+    public static final Integer DEFAULT_MAX_RETRIES = 3;
```
breedx-splk, line 14: nit: prefer primitives for simple things.

signalfx-java/src/main/java/com/signalfx/metrics/connection/HttpDataPointProtobufReceiverFactory.java
```diff
@@ -42,6 +48,11 @@ public HttpDataPointProtobufReceiverFactory setMaxRetries(int maxRetries) {
         return this;
     }
 
+    public HttpDataPointProtobufReceiverFactory setNonRetryableExceptions(List<Class<? extends IOException>> clazzes) {
+        this.nonRetryableExceptions = clazzes;
```
breedx-splk, line 52: nit: prefer copying the list contents rather than holding reference (to prevent future mutability).

signalfx-java/src/test/java/com/signalfx/metrics/connection/HttpDataPointProtobufReceiverConnectionTest.java
Excaleo: Na you're right, I opted to use `await` with the timeout instead, I've changed the test to schedule the release of the latch after a certain time period instead.

It's quite tricky with the retry test case, we have to release the latch after the retried request since they share the same latch. Currently I've scheduled the latch to `countDown()` after 3s. 3s was chosen because it's sufficient for the retry to trigger. Anything less and it might exit before the retry.

breedx-splk: Did we just turn a 2 second test into a 3 second test? 😁 
The coordination between threads for testing purposes can be a hassle, indeed. 

Maybe we can rethink this -- Seems like there are two testing goals here: 
1. to test the case where we *DO* retry when it's a retryable exception (not in the non-retryable list) 
2. to test the case were we *DO NOT* retry when the exception is in the non-retryable list.

For case 1, rather than depending on `SocketTimeoutException`, can we use a different exception? Something that fails almost immediately. I'm not sure offhand but there's gotta be something, right? Maybe remove `UnknownHostException` from the list for this test and force that?

For case 2, I think it should just be as simple as forcing that `UnknownHostException` with the default list, right?

Excaleo: > For case 1, rather than depending on SocketTimeoutException, can we use a different exception? Something that fails almost immediately. I'm not sure offhand but there's gotta be something, right? Maybe remove UnknownHostException from the list for this test and force that?

While we can force a retry with UnknownHostException, it would be hard to assert that the retry has indeed taken place without some sort of mock server keeping track of request counts, or some sort of http client library to keep track of request counts (afaik we can't modify this for the DatapointPointReceiver).

Throwing an exception in the server/handler doesn't necessarily propagate back to the client either.

Let me see if there are any other retry scenarios that I can force without relying on socket timeouts.

Excaleo: For now I've tuned the test case to trigger the latch after 500ms instead of 3s (by tuning the receiver timeout).

breedx-splk: Well, `HttpDataPointProtobufReceiverConnectionTest` failed in ci/cd. Womp womp.

breedx-splk: I think this will have to do for now. Thanks for sticking it out, I appreciate it.

breedx-splk (approved):
Thanks for the contribution! 

@Excaleo While we do appreciate that people are still using this library and we do support it for now, our focus is definitely on OpenTelemetry, and we would encourage users of this library to consider migrating. Thanks!

signalfx-java/src/main/java/com/signalfx/connection/RetryDefaults.java
```diff
@@ -0,0 +1,19 @@
...
+
+public final class RetryDefaults {
+    private RetryDefaults() {
+    }
+
+    public static final int DEFAULT_MAX_RETRIES = 3;
+    public static final List<Class<? extends IOException>> DEFAULT_NON_RETRYABLE_EXCEPTIONS = Arrays.asList(
+            InterruptedIOException.class,
+            UnknownHostException.class,
+            ConnectException.class);
```
mateuszrzeszutek, line 18: Nit
```suggestion
    public static final List<Class<? extends IOException>> DEFAULT_NON_RETRYABLE_EXCEPTIONS = Collections.unmodifiableList(Arrays.asList(
            InterruptedIOException.class,
            UnknownHostException.class,
            ConnectException.class));
```

signalfx-java/src/main/java/com/signalfx/metrics/connection/HttpDataPointProtobufReceiverFactory.java
```diff
@@ -42,6 +49,11 @@ public HttpDataPointProtobufReceiverFactory setMaxRetries(int maxRetries) {
         return this;
     }
 
+    public HttpDataPointProtobufReceiverFactory setNonRetryableExceptions(List<Class<? extends IOException>> clazzes) {
+        this.nonRetryableExceptions = Collections.unmodifiableList(clazzes);
```
mateuszrzeszutek, line 53: `unmodifiableList()` doesn't copy the list, it just creates a thin immutable wrapper over it.
```suggestion
        this.nonRetryableExceptions = new ArrayList<>(clazzes);
```

mateuszrzeszutek (approved):
Thanks for contributing @Excaleo !
I left a couple of minor nitpicking comments, but overall it looks good to me 👍

signalfx-java/src/main/java/com/signalfx/metrics/connection/HttpDataPointProtobufReceiverFactory.java
Excaleo: Ah ok, thanks for that insight, I might do `Collections.unmodifiableList(new ArrayList<>(clazzes))` instead then, this will create an unmodifiable view without any reference to the original ArrayList.

Excaleo:
Thanks for the reviews @breedx-splk and @mateuszrzeszutek ! 

We are aware that the focus is on Opentelemetry at the moment, we're just using this library for compatibility with our existing stacks. Hopefully, in the future we'll be leveraging the Opentelemetry libraries instead!

I've added one last commit to address the immutability comments.

mateuszrzeszutek (approved):

mateuszrzeszutek:
Thanks @Excaleo ! 🚀

Excaleo:
Hey @mateuszrzeszutek @breedx-splk , just wondering when the next release would be to roll out these changes?

mateuszrzeszutek:
Hey @Excaleo , 
I'll start the release process -- you can expect the 1.0.24 version to appear either today or early next week.

PR #271, opened 2022-08-30
Prepare 1.0.23 release

mateuszrzeszutek:


pom.xml
```diff
@@ -24,7 +24,7 @@
     <os72.protobuf-shaded.plugin-version>${protobuf.version}</os72.protobuf-shaded.plugin-version>
     <os72.protobuf-shaded.version>3-11-1</os72.protobuf-shaded.version>
     <protobuf.version>3.11.1</protobuf.version>
-    <slf4j.version>2.0.0</slf4j.version>
+    <slf4j.version>1.7.36</slf4j.version>
```
mateuszrzeszutek, line 27: I've decided to revert the slf4j change for the time being -- this lib is used in the javaagent, and right now we're not sure if we can safely bump the version.

breedx-splk: 😿

breedx-splk (approved):

PR #245, opened 2022-07-08
Add a retry strategy with max retries

irisgve:
We encounter a bunch of errors accessing the sfx endpoints that are potentially retryable and we want to make this property configurable.

PR #321, opened 2023-03-14
Enable retries on 429 in retry handler

Excaleo:
## Context
In #245, a retry strategy that copied the behaviour of the golang multitoken sink implementation was added. This PR addresses a bug within the strategy. In the golang multitoken sink, a `429` response would actually result in a status code of `-1` in the golang struct for storing status codes, resulting in a retry. I can elaborate on the implementation if you would like that context.

In the current implementation, the `429` is not mapped to `-1`, and will therefore not result in a retry.

## Changes

- Add explicit retry on 429s


Excaleo:
After checking the test cases it seems that 429 is returned as the status code and explicitly not retried.

PR #322, opened 2023-03-14
Fix license file included in built jars

mateuszrzeszutek:


breedx-splk (approved):
Thanks for banging this out!

PR #332, opened 2023-05-04
Split out signalflow into separate module

breedx-splk:
This modularizes the code quite a bit. There are some tangled dependencies, so 2 additional modules also needed to be factored out: `signalfx-connection` and `signalfx-endpoint`.

The dependency graph is basically:

```mermaid
graph TD;
    signalfx-connection-->signalfx-endpoint;
    signalfx-signalflow-->signalfx-endpoint;
    signalfx-signalflow-->signalfx-connection;
    signalfx-java-->signalfx-connection;
    signalfx-java-->signalfx-endpoint;
    signalfx-java-->signalfx-signalflow;
```

There was one code change needed to eliminate cycles there. The `SignalFxMetricsException` which was an unnecessary (named/tag) subclass of `RuntimeException` was replaced with just `RuntimeException`.

While I do think this split helps to better organize and show where dependencies are and to eventually pull out signalflow, it also highlights that signalflow still has some dependencies on this common stuff.


signalfx-connection/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+    </dependency>
+    <!-- test -->
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
+    </dependency>
```
mateuszrzeszutek, line 66: That's defined as a compile dependency in the parent pom

signalfx-connection/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+  <parent>
+    <groupId>com.signalfx.public</groupId>
+    <artifactId>clients-parent</artifactId>
+    <version>1.0.30</version>
+  </parent>
+
+  <properties>
+    <!-- This prevents shading from happening in this module -->
+    <skipShaded>true</skipShaded>
+    <maven.deploy.skip>false</maven.deploy.skip>
```
mateuszrzeszutek, line 16: Just wondering, what's this for?

pom.xml
```diff
@@ -176,11 +177,11 @@
...
-      <dependency>
-        <groupId>org.eclipse.jetty.websocket</groupId>
-        <artifactId>websocket-client</artifactId>
-        <version>${jetty.version}</version>
-      </dependency>
+<!--      <dependency>-->
+<!--        <groupId>org.eclipse.jetty.websocket</groupId>-->
+<!--        <artifactId>websocket-client</artifactId>-->
+<!--        <version>${jetty.version}</version>-->
+<!--      </dependency>-->
```
mateuszrzeszutek, line 184: To be removed?
```suggestion
```

signalfx-signalflow/pom.xml
```diff
@@ -0,0 +1,110 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty.websocket</groupId>
+      <artifactId>websocket-client</artifactId>
+      <version>${jetty.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-server</artifactId>
+    </dependency>
```
mateuszrzeszutek, line 102: That's a test dependency actually

signalfx-java/pom.xml
```diff
@@ -43,15 +43,37 @@
...
+      <groupId>com.signalfx.public</groupId>
+      <artifactId>signalfx-connection</artifactId>
+      <version>${version}</version>
+    </dependency>
     <dependency>
-      <groupId>org.eclipse.jetty.websocket</groupId>
-      <artifactId>websocket-client</artifactId>
+      <groupId>com.signalfx.public</groupId>
+      <artifactId>signalfx-endpoint</artifactId>
+      <version>${version}</version>
```
mateuszrzeszutek, line 55: These two also appear in the `<!-- compile -->` section; in an identical form.

mateuszrzeszutek (commented):
Awesome!

signalfx-connection/pom.xml
breedx-splk: I think the parent just sets up the versions...so this is definitely required, I confirmed that removing it breaks the build.

signalfx-connection/pom.xml
breedx-splk: It's wonky.  So without this (and the new setting in the parent pom), every child module/jar gets its dependencies shaded, which can result in public/external APIs having breaking changes to their interface. So a method that normally takes something like `org.apache.FooImpl` gets changed to `com.signalfx.shaded.org.apache.FooImpl` and this breaks all the callers. I couldn't find another way around that, unfortunately.

signalfx-java/pom.xml
breedx-splk: Thanks!

breedx-splk:
ready for another look @mateuszrzeszutek thanks!

signalfx-connection/pom.xml
mateuszrzeszutek: OOF
maven magic 🪄

signalfx-connection/pom.xml
mateuszrzeszutek: > I think the parent just sets up the versions...so this is definitely required, I confirmed that removing it breaks the build.

I'm not saying that it's not required; I'm saying that if it's a test dep it needs `<scope>test</scope>`

breedx-splk: Yeah, that doesn't work because it's used by the `RetryStrategy` and `RetryHandler` which are production classes.

mateuszrzeszutek (approved):
🎉

PR #337, opened 2023-05-15
Remove v1 metrics sender

laurit:
`/v1/metrics` endpoint used by this code responds with 404. Open question is what to do when user sets version to 1, this pr ignores this, but could also throw either when setting the version or when exporting.


mateuszrzeszutek (approved):

laurit:
> > Open question is what to do when user sets version to 1, this pr ignores this, but could also throw either when setting the version or when exporting.
> 
> We should fail as early as possible -- or not allow the user to set the version in the first place.

we could remove the `setVersion` method or make it a no-op, or we could throw an `IllegalArgumentException` when it is called with `1` or any other value besides `2`

mateuszrzeszutek:
Hm, let's deprecate it and make it a noop then -- seems to be the safest choice

mateuszrzeszutek (approved):

laurit:
> Hold up. The description says `/v1/metrics` but the implementation has `/v1/datapoint`. Are we confident that's no longer available? I'm all about removing support for it, I just want to be extra sure we're doing what we intended.

I initially also thought that it used `/v1/datapoint` because of https://github.com/signalfx/signalfx-java/blob/f09cb84b3bce00f871146086bc91a1c54689d43b/signalfx-java/src/main/java/com/signalfx/metrics/connection/HttpDataPointProtobufReceiverConnection.java#L58 but it really uses `/v1/metrics`
https://github.com/signalfx/signalfx-java/blob/f09cb84b3bce00f871146086bc91a1c54689d43b/signalfx-java/src/main/java/com/signalfx/metrics/connection/HttpDataPointProtobufReceiverConnection.java#L93

PR #343, opened 2023-05-18
Move metrics code from signalfx-java to signalfx-metrics

laurit:
Move all code from `signalfx-java` to `signalfx-metrics`. Additionally removes `jackson-databind` dependency from `signalfx-metrics` and replaces it with `jackson-jr-objects`. Currently new metrics module is not shaded, we need to decide whether to shade it or publish both shaded and unshaded versions. For embedding to agent unshaded would give us less duplicate dependencies.


signalfx-metrics/pom.xml
```diff
@@ -0,0 +1,123 @@
...
+  <parent>
+    <groupId>com.signalfx.public</groupId>
+    <artifactId>clients-parent</artifactId>
+    <version>1.0.31</version>
+  </parent>
+
+  <properties>
+    <!-- This prevents shading from happening in this module -->
+    <skipShaded>true</skipShaded>
+    <maven.deploy.skip>false</maven.deploy.skip>
```
laurit, line 15: Copied this from another module, no idea why we would need to set it

signalfx-metrics/src/main/java/com/signalfx/metrics/aws/AWSInstanceInfo.java
```diff
@@ -41,18 +43,20 @@ public static String get(int timeoutInMs) {
...
         } catch (Exception e) {
             log.trace("Exception trying to execute {}, Exception: {} ", request, e);
         }
 
         return null;
     }
 
+    @VisibleForTesting
+    static String parse(InputStream inputStream) throws Exception {
+        Map<String, Object> result = JSON.std.mapFrom(new InputStreamReader(inputStream, StandardCharsets.UTF_8));
```
mateuszrzeszutek, line 56: Shouldn't we close the `inputStream`?

mateuszrzeszutek (commented):
Looks great 👍 

The build's still failing on some duplicated files though

signalfx-metrics/src/main/java/com/signalfx/metrics/aws/AWSInstanceInfo.java
laurit: Thanks, I'll change the caller to close the stream.

laurit:
> The build's still failing on some duplicated files though

Should be fixed now.

mateuszrzeszutek (approved):

PR #347, opened 2023-05-19
Remove commons-io dependency from metrics

laurit:


signalfx-metrics/src/main/java/com/signalfx/metrics/SourceNameHelper.java
```diff
@@ -49,6 +52,9 @@ public static String getDefaultSourceName() {
 
     @Deprecated
     public static String getAwsInstanceId() throws IOException {
```
laurit, line 54: Alternatively could just delete this method, it has been deprecated since 2016

mateuszrzeszutek: This is fine to me 🤷

mateuszrzeszutek (approved):

signalfx-metrics/src/main/java/com/signalfx/metrics/SourceNameHelper.java
laurit: deleted it

PR #361, opened 2023-06-12
Replace protobuf dependency

laurit:
Replace `com.github.os72:protobuf-java-shaded-3.11.1` with `com.google.protobuf:protobuf-java`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,10 +21,8 @@
     <jackson.version>2.15.2</jackson.version>
     <jackson-databind.version>2.15.2</jackson-databind.version>
     <jetty.version>9.4.51.v20230217</jetty.version>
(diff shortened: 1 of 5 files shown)


breedx-splk:
I have no idea about the original intention of using the os72 flavor, but it _appears_ that the original authors might have wanted consistency (eg. one protobuf vendor across the tools). This change doesn't completely eliminate the os72 stuff, because it's still used for the `protoc-jar-maven-plugin` I guess. Is there a non-os72 equivalent? I didn't see one at first glance...

Can you please give a little justification/intent behind this change in the description?

breedx-splk:
Ok cool, seems worth it to get rid of the duplicated classes for sure. Thanks for clearing that up.

breedx-splk (approved):
Let's give it a shot.

mateuszrzeszutek (approved):

PR #401, opened 2023-10-27
Make classes final to discourage type inheritance from these.

breedx-splk:


laurit:
As we plan to delete these classes perhaps it would make sense to avoid doing any changes that aren't necessary.

breedx-splk:
Some things weren't meant to be.

PR #414, opened 2023-12-14
Support timezone param for WebSocketTransport execute

nishantagarwal1990:
The SignaFlow client for java does not have a way to specify the timezone while the signalfx-python lib does. We would like to have similar capabilities.

### DIFF
--- a/signalfx-signalflow/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
+++ b/signalfx-signalflow/src/main/java/com/signalfx/signalflow/SignalFlowClient.java
@@ -56,7 +56,7 @@ public class SignalFlowClient implements AutoCloseable {
 
     /**
      * This method is deprecated and will be removed in the next major release. Use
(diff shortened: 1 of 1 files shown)


breedx-splk (approved):
This seems fine to me, but I'd like to see two things:

1. Can you add a link to the docs for the timezone parameter? Is is just passed over the wire to the receiving side or is this actually configuring something in the transport itself? 
2. How would a user know what kind of value to pass in there? Do they pass in `PDT` or do they pass in `America/Los_Angeles` or would either work?

nishantagarwal1990:
The timezone is passed over the wire to the receiving side as a query parameter. The docs link is https://dev.splunk.com/observability/docs/signalflow/#Supported-SignalFlow-time-zones

The supported timezones are https://dev.splunk.com/observability/docs/time_zone

breedx-splk:
Sorry what I meant was can you please add that in the javadocs so that a user can see it? 👍🏻

PR #449, opened 2024-04-18
Upgrade httpcomponents

breedx-splk:
So the 4.x branch of httpclient hasn't seen a release since Nov. 2022. This is old dependency is causing vulnerabilities to be detected, by way of an old version of`commons-codec` ([see here](https://github.com/apache/httpcomponents-client/blob/4.5.x/pom.xml#L79)).

This is a breaking change for any user that might have been using the (public) `RetryStrategy` class directly. Most likely that class should not have been public to begin with.


signalfx-metrics/src/main/java/com/signalfx/metrics/connection/HttpClientConnectionManagerFactory.java
```diff
@@ -39,7 +41,7 @@ private static class SSLConnectionSocketFactoryWithTimeout extends SSLConnection
     private final int timeoutMs;
 
     public SSLConnectionSocketFactoryWithTimeout(int timeoutMs) {
-      super(SSLContexts.createDefault(), SSLConnectionSocketFactory.STRICT_HOSTNAME_VERIFIER);
```
breedx-splk, line 42: I couldn't find an equivalent of this in httpclient5. Could use a recommendation.

breedx-splk:
Well, it's here if we ever want to revisit.

